PR #24, opened 2015-02-21
feat(anim): Add animator service

zewa666:
feat(anim): Add animator service

Adds an animator service which is meant to provide basic animation
support via CSS classes

### DIFF
--- a/src/animator.js
+++ b/src/animator.js
@@ -1,8 +1,8 @@
 export class Animator {
   constructor() {
+    Animator.instance = this;
     this.animationStack = [];
   }
-
 
   addMultipleEventListener(el, s, fn) {
     var evts = s.split(' ');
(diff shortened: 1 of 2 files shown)


zewa666:
For the next steps it would be useful to merge this into templating and wire up with the Framework, as we spoke yesterday via Gitter. After that the next phase is to start work on the AttachedBehaviors If,Show, Repeater, ...

zewa666:
@stoffeastrom made a good point about providing the Animation-Support via opt-in plugins. So I've updated the pull request to include a basic animator which essentially just returns immediately resolved Promsies. I think we should then follow the idea to provide concrete implementations via plugins. The first plugin could be the previously created CSS-Animator.

EisenbergEffect:
@zewa666 I suspect we will need to make some more modifications to the view slot implementation in the future, but I think this is good for now. I'm excited about getting the animation working. I've got this merged. I'll hook up the work in the framework to make registration happen, per our discussion. This will go out in the next release. Should I create a new repo for the animation implementation? What should we call it? animator-css?

PR #28, opened 2015-03-01
feat(OptionsProperty) property changed handling for dynamic properties

### LINKED ISSUE #27: support property change handling in dynamic options properties

jdanyow:


jdanyow:
no tests but I did do an integration test in the skeleton and confirmed `dynamicPropertyChanged` was invoked.

PWKad:
@jdanyow Not including a test that the event is fired means in future it is easily possible that this feature gets broken silently - what scenarios should it get fired?  Perhaps someone can add the test before it gets merged in?  I can try if you can define the scenarios.

jdanyow:
@PWKad you caught me!  :smile: 

I encountered this issue while working on https://github.com/aurelia/binding/issues/38 and discussed the changes with @EisenbergEffect before creating the PR.  

Maybe we can brainstorm on the best way to add tests for this- it's a tricky one because the change is in templating but to confirm it's working you kind of need binding and maybe even templating-resources.  There's probably an easier way that I'm not seeing.

jdanyow:
forgot to explain the scenario:

This PR allows your behavior to be notified when you do something like [this](https://github.com/aurelia/templating-resources/blob/master/src/global-behavior.js#L8) and a dynamic property changes.

PWKad:
Got it, was just curious because I would hate to have it lost in history because no test was there to back it up :)

I am trying to out together some automated tests now so I can look for a way to incorporate it there or in the e2e tests?

PR #39, opened 2015-03-19
fix(view-slot): add firstChild null checks

### LINKED ISSUE #34: ViewSlot firstChild null not being checked
i'm getting an exception when i run a `todo` sample.  it looks like `ViewSlot` is not checking if .firstChild is null before referencing its properties.

[typescript repo containing sample](https://github.com/cmichaelgraham/aurelia-typescript)
[todo project within repo](https://github.com/cmichaelgraham/aurelia-typescript/tree/master/todo-require-vs-ts)
[running code illustrating exception](http://cmichaelgraham.github.io/aurelia-typescript/todo-require-vs-ts/index-latest.html)
[no check for firstChild null](https://github.com/aurelia/templating/blob/master/src/view-slot.js#L142)

cmichaelgraham:
add null checks for case when firstChild is null.  this occurs when the aurelia-app attribute is in a child element of body instead of on the body element, for example

Fixes #34

### DIFF
--- a/src/view-slot.js
+++ b/src/view-slot.js
@@ -72,8 +72,7 @@ export class ViewSlot {
       view.attached();
       // Animate page itself
       var element = view.firstChild ? view.firstChild.nextElementSibling : null;
-      if(view.firstChild !== undefined &&
-        view.firstChild !== null &&
+      if(view.firstChild !== null &&
         view.firstChild.nodeType === 8 &&
         element !== null &&
         element.nodeType === 1 &&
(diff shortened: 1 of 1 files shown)


PWKad:
:beers: for @cmichaelgraham

EisenbergEffect:
I'm just trying to find out if there's actually a case where firstChild could be undefined rather than null. This is perf sensitive code so if we can get away with only one check that would be better.

EisenbergEffect:
Reading up on firstChild, I think it can be changed to `view.firstChild !== null` only.

cmichaelgraham:
ok, should i go ahead and redo the PR?

EisenbergEffect:
You should be able to make the change and commit/push to your branch. I don't think you need to do a new PR. It should get updated. Just let me know when it's pushed.

EisenbergEffect:
@cmichaelgraham One final thing is required before I can accept this. You need to sign the CLA: https://docs.google.com/forms/d/1Npd68d32UDUKUYuCDDrW0VN5Lv8p-a6Ziaa2JFKp6Cg/viewform?c=0&w=1 I thought you had signed it, but I don't see your response in my list.

cmichaelgraham:
done & done !! :)

Issue #8, opened 2015-01-29
Compile linked stylesheets of shadowDOM custom elements


olanod:
When I have a `<link rel='stylesheet' href='...' />` inside a template of a custom element using shadowDOM I expect it to be fetched and in-lined inside the shadow root just like polymer does.

EisenbergEffect:
That is a good idea.

EisenbergEffect:
Can you show me the markup you want to write and then also show me what you want the output to be? I think I understand, but want to be sure.

olanod:
Something like this:
```
<template>
    <link rel='stylesheet' href='...' />
    <h1>Hello World</h1>
</template>
```
And produces:
```
#shadow-root
    <style>
    ....
    </style>
    <h1>Hello World</h1>
```

EisenbergEffect:
Gotcha. I think we can make that happen ;)

Hendrixer:
:+1:

yelvert:
Wonder if this could be automatically included, at least in Behaviors? Something like `Behavior.useStyle(optionalPath)`, and that loads and attaches the stylesheet, same as the template.

EisenbergEffect:
The `require` element now supports this. You can import css files with require. if you want them to be scoped to the shadow dom, then you do it like this:

```
<require from="./my.css" as="scoped"></require>
```

PR #42, opened 2015-03-26
Shadowdom inline style

olanod:
Fixes https://github.com/aurelia/templating/issues/8 but it's not done yet, it's still a concept because I don't know if it's the aurelia way of doing it and because I haven't been able to test it(some help here?) but the idea is clear.  
I think it's the best option for style encapsulation and the `inlineLinkedCSS` function could be use for the view bundling, inlining the styles when the html is compiled for production.

### DIFF
--- a/config.js
+++ b/config.js
@@ -18,6 +18,7 @@ System.config({
     "aurelia-path": "github:aurelia/path@0.5.0",
     "aurelia-task-queue": "github:aurelia/task-queue@0.3.0",
     "core-js": "github:zloirock/core-js@0.8.1",
+    "text": "github:systemjs/plugin-text@0.0.2",
     "github:aurelia/binding@0.5.0": {
       "aurelia-dependency-injection": "github:aurelia/dependency-injection@0.6.0",
       "aurelia-metadata": "github:aurelia/metadata@0.4.0",
@@ -40,4 +41,3 @@ System.config({
     }
   }
 });
-
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
       "aurelia-metadata": "github:aurelia/metadata@^0.4.0",
       "aurelia-path": "github:aurelia/path@^0.5.0",
       "aurelia-task-queue": "github:aurelia/task-queue@^0.3.0",
-      "core-js": "github:zloirock/core-js@^0.8.1"
+      "core-js": "github:zloirock/core-js@^0.8.1",
+      "text": "github:systemjs/plugin-text@^0.0.2"
     }
   },
   "devDependencies": {
--- a/src/html-behavior.js
+++ b/src/html-behavior.js
@@ -4,7 +4,7 @@ import {TaskQueue} from 'aurelia-task-queue';
 import {ViewStrategy} from './view-strategy';
 import {ViewEngine} from './view-engine';
 import {ContentSelector} from './content-selector';
-import {hyphenate} from './util';
+import {hyphenate, swapLinksWithStyleContent} from './util';
 import {BindableProperty} from './bindable-property';
 import {BehaviorInstance} from './behavior-instance';
 
@@ -109,13 +109,22 @@ export class HtmlBehaviorResource extends ResourceType {
         viewStrategy.moduleId = Origin.get(target).moduleId;
       }
 
-      return viewStrategy.loadViewFactory(container.get(ViewEngine), options).then(viewFactory => {
-        if(!transientView){
-          this.viewFactory = viewFactory;
-        }
+      return viewStrategy.loadViewFactory(container.get(ViewEngine), options)
+        .then(viewFactory => {
+          if(!transientView){
+            this.viewFactory = viewFactory;
+          }
 
-        return viewFactory;
-      });
+          return viewFactory;
+        })
+        // inline stylesheets on shadowDOM elements
+        .then( viewFactory => {
+          if(this.targetShadowDOM){
+            return swapLinksWithStyleContent(viewFactory.template)
+              .then(template => viewFactory);
+          }
+          return viewFactory;
+        });
     }
 
     return Promise.resolve(this);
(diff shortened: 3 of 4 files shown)


src/custom-element.js
```diff
@@ -76,6 +76,10 @@ export class CustomElement extends ResourceType {
       instruction.contentFactory = compiler.compile(fragment, resources);
     }
 
+    // inline linked css
+    if (this.userShadowDOM)
```
bryanrsmith: Typo: "usesShadowDOM"

olanod:
I'll check how to load the stylesheets from the load method, my doubt was if I should use directly `System.import('style.css!text')` or if aurelia has a resource loader(I think it does?) that handles the path resolution etc.  
I think what we want to use is the text plugin of system.js, so it has to be included as a dependency somewhere. The css plugin imports the stylesheets and adds them to the `head` element but in this case we want the content of the file to inline it in the template.  
This is what I do in my custom elements for now(before I had `beforeCompile`) with the text plugin and it works fine:

    import {addStyleToShadowElement as addStyle} from 'lib/util';
    import style from './some-element.css!text';

    export class SomeElement {
        static metadata() {
            return Behavior
                .customElement('some-element')
                .useShadowDOM();
        }
        static inject() { return [Element] }
        constructor(ele) {
            this.element = ele;
        }

        bind() {
            addStyle(this.element, style);
        }
    }

(lib/util)

    export function addStyleToShadowElement(element, style) {
        if (element.shadowRoot) {
            let styleSheet = document.createElement('style');
            styleSheet.appendChild(document.createTextNode(style));
            element.shadowRoot.appendChild(styleSheet);
        }
    }

olanod:
I forgot, can you also help me to setup a test for this? I see this repo doesn't include the implementation of `BindingLanguage` interface which is needed to compile the element so it should be included as a devDependency(available now in jspm 1.5) is that right?

olanod:
@EisenbergEffect I force-updated this PR based on the newer version of the code.
I still haven't test it(I really tried :disappointed:) but tell me if the concept is right. 
I put the code in the `load` method where resources should be loaded(right?) and It just modifies the template in the viewFactory.

EisenbergEffect:
You are on the right track. What I will want to do is remove the reference to System.js and instead add more functionality onto our loader abstraction for loading css. Then internally, we can use the text plugin. This will allow us to support both system.js and require.js which have different syntax. It will also let us abstract this along with other pieces so that those who need to use a custom loader and handle it all in one place and not have to worry about the templating engine having something like this hard coded.

For the next version, I can extend our loader abstraction with a mechanism we can use here.
Unit tests aside, were you able to try this out in an application and see if it worked? One thing we might want to do is investigate scope stye polyfills. Maybe we can achieve a similar effect if there's no native scoping? Since we control the css loading, we could also make transformations if needed. Thoughts?

olanod:
So I should wait until the loader implements the text plugin right?  
What about letting users load the styles themselves, now that decorators are used a lot  we could have a decorator for that
```
import style from './style.css!text';
import {withStyle} from 'aurelia-templating';

@withStyle(style)
class MyElement {
}
```
The decorator could work like this
```
export function withStyle(style) {
    return function(target) {
        // do something first if its already defined
        target.beforeCompile = function(template) {
             addStyleToTemplate(template, style);
        }
    }
}
```

EisenbergEffect:
Yes, I will try to add the text loading to the loader for the next release. Regarding using decorators...you could do it, but in a different way, by tapping into the HtmlBehavior load callback. Perhaps we can make that more extensible?

Aaike:
I want to note that this technique only works for custom elements and viewmodels that have a view.
It would be good if this also works when using the @noView decorator.

also i slightly changed to way the code looks so it doesn't need an import statement.
it would work a lot like the `@useView` by taking a parameter to a moduleName

So i'm doing this
```
constructor(element){
    this.element = element;
    injectStyle(this.element,'prism/themes/prism-okaidia.css');
}
```

and this is the function i'm using
```
export function injectStyle(el, stylePath) {
  System.import(stylePath+"!text").then(data=>{
    let style = document.createElement('style');
    style.appendChild(document.createTextNode(data));
    el.appendChild(style);
  });
}
```

This would work well with a @withStyle or @useStyle decorator like this :
```
@useStyle('moduleName/style.css')
```

Aaike:
I also want to note that this technique only works for custom elements and viewmodels that have a view. It would be good if this also works when using the @noView decorator.
for now i have to do something like this :

```
constructor(element){
    this.element = element;
    this.root = element.createShadowRoot();
    injectStyle(this.root,'moduleName/style.css');
}
```

would be great if 
`@useShadowDOM` used in combination with `@useView` and `@useStyle` would do exactly what the code above does.

Aaike:
[here is a plunkr](http://plnkr.co/edit/1hcTDeFEeN46B9PzY3II?p=preview) with a simple custom element that would fall under this use case for what i'm trying to do with the @noView , @useShadowDOM and @useStyle combination.

check out the `markdown.js` in that plunk. 
I've commented the decorators that would ideally replace what i'm currently doing on line 16 and 21.

Also as i see it it should work with all scenarios, with or without shadow dom and with or without a view template.

Aaike:
a solution for this can now be tracked here : https://github.com/aurelia/templating/issues/65

EisenbergEffect:
FYI Some new extension points coming soon to enable this. Stay tuned. It should land in the next week or two.

EisenbergEffect:
Good new. In the next release, you will be able to do this using our `<require>` element.

To load global styles with a view:

```html
<require from="./my-element.css"></require>
```

To load scoped/shadow dom styles for a view

```html
<require from="./my-element.css" as="scoped"></require>
```

This will also work with our bundling. So, we will be able to bundle all the css resources and load them from the bundle transparently. It also enables sharing of css across shadow dom boundaries. If two different views request the same css file, that's only one import.

PR #53, opened 2015-04-10
fix(BindableProperty): null changeHandler

jdanyow:
hit an error here because changeHandler was null (instead of undefined) when trying some things with the `on-attached.call` custom attribute we were playing around with in the gitter.

### DIFF
--- a/src/bindable-property.js
+++ b/src/bindable-property.js
@@ -62,7 +62,7 @@ export class BindableProperty {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Reviewed. I ended up fixing it a bit differently, but it should be good to go for the next release.

jdanyow:
:+1: nicer the way you did it.

PR #56, opened 2015-04-14
feat(templating): remove dependency cycle

cmichaelgraham:
remove cycle between view-engine, html-behavior, and module-analyzer by
combining into view-engine.

also update references in decorators, index, and composition-engine

this helps amd/requirejs scenarios where cycles produce undefined imports


cmichaelgraham:
please hold off on this.  @colindembovsky has a better solution he's working on :+1:

EisenbergEffect:
Ok, I'm going to close this until we get the "real" fix :)

colindembovsky:
@EisenbergEffect @cmichaelgraham Seems like it's a bit harder than I thought - all I've managed to do is increase the dependency cycle from 3 files to 4 files :smile:. We may need some help to remove the cycle, since it's not a simple fix.

PR #59, opened 2015-04-19
fix(html-behavior): fix cycle issues

cmichaelgraham:
change the import strategy to fix cycle issue


cmichaelgraham:
lets discuss this before you apply it.

i believe that the new version of babel produces code that allows the cycle to work correctly using amd modules, making this PR unnecessary.

cmichaelgraham:
should no longer be necessary with new babel compiler

PR #71, opened 2015-05-07
fix(view-slot): add fallback for missing nextElementSibling

martingust:
When the templates get loaded nextElementSibling is missing in IE. Added helper method to search through nextSiblings to get the next element.

Closes aurelia/animator-css#8


zewa666:
Looking forward to test this, great work @martingust :+1:

PR #74, opened 2015-05-14
feat(customAttribute): enable defaultBindingMode for customAttribute

manuel-guilbault:
Add a second parameter to customAttribute decorator, that allow to
specify the default binding mode on a custom attribute.

### DIFF
--- a/src/decorators.js
+++ b/src/decorators.js
@@ -28,10 +28,11 @@ export function customElement(name){
 
 Decorators.configure.parameterizedDecorator('customElement', customElement);
 
-export function customAttribute(name){
+export function customAttribute(name, defaultBindingMode){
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
I need to think over this. It seems like a great idea though. Props also for the high quality PR with tests :)

manuel-guilbault:
Thanks :)

Actually, the idea came from a custom attribute I'm working on. I want to create a "focus" attribute, similar to the focus binding in Durandal. And without this new feature, developers would be forced to use the ".twoWay" command instead of the common ".bind" command.

EisenbergEffect:
Thanks again. This was a great quality PR! You should contribute more :)

PR #87, opened 2015-06-05
feat(anim): add animation events const

zewa666:
adds a constant providing all possible events to be raised by the concrete Animator implementations

part of issue https://github.com/aurelia/animator-css/issues/13

### DIFF
--- a/src/animation-event.js
+++ b/src/animation-event.js
@@ -1,4 +1,4 @@
-export const bindingMode = {
+export const animationEvent = {
   enterBegin:   'animation:enter:begin',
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@zewa666 That should not be named `bindingMode`. I think you might have made a copy/paste error there.

EisenbergEffect:
I think we are trying to keep everything singular as well. So probably name the file `animation-event.js` and the object `animationEvent`.

zewa666:
oh god what a stupid mistake :) sry it's fixed now

Issue #88, opened 2015-06-05
Add base animator methods for integrating with JS based animators.


Aaike:
A couple of functions should be added to the base animator to allow full use of JS based animators and some extra methods that make sense for CSS based animators as well.
```javascript
animate(element,className,options)
```

This can trigger a single event or callout like "bounce" or "shake".
For css animators this is a class that will be added to the element and removed right after the animation is done.
For JS based implementations the `className` is used as the `effectName` or as an object that specifies the properties to be animated.

```javascript
runSequence(sequence)
```

The sequence is an array of `effectNames` of `classNames` that represent animations to be animated with the `animate` method one after another.

```javascript
registerEffect(effectName,properties)
unregisterEffect(effectName)
```

Register and unregister an effect with the animator (for JS based animators)

Aaike:
An implementation of a JS based animator using velocity can be seen here : https://github.com/gooy/aurelia-animator-velocity/

zewa666:
@EisenbergEffect as mentioned in the PR issue, that one looks great and is ok to merge :+1: 
I'd also like to highlight @Aaike's implementation of the VelocityAnimator, very nice work and adds now even more flexibility to the Aurelia-EcoSystem. Great work !!!

Aaike:
A basic implementation and demo for GSAP Animator can now be found at : https://github.com/gooy/aurelia-animator-greensock

EisenbergEffect:
Sweet! I think a guest blog post is in order soon :)

Aaike:
last and perhaps also least this time, the TinyAnimate Animator : https://github.com/gooy/aurelia-animator-tinyanimate

EisenbergEffect:
Nice! Very nice work. Do you think you'll be able to address the browser issues that people in the gitter have discovered?

Aaike:
thx ! Yes i have a pretty good idea on how to fix them ;)

However one of those issues is because of Aurelia i think, and that is when switching pages in IE11 : https://github.com/aurelia/templating/issues/86

Aaike:
One more thing
when using the animator somewhere, one would have to import it like this:
`import {VelocityAnimator} from "blabla"`
or as
`import {CSSAnimator} from "blabla"`
and then inject it.

So you would have either `VelocityAnimator` or `CssAnimator` references in all your pages and VM's etc...
Which makes it not so easy to swap one with the other..

is there a way we can avoid declaring exactly what animator implementation to use in our VMs ?

[...]

PR #89, opened 2015-06-05
feat(animator): add extra animator methods

Aaike:
add extra animator methods for extra functionality and for working with
JS based animators

fixes: https://github.com/aurelia/templating/issues/88

### DIFF
--- a/src/animator.js
+++ b/src/animator.js
@@ -7,19 +7,91 @@ export class Animator {
     return Promise.resolve(false);
   }
 
+  /**
+   * Execute an 'enter' animation on an element
+   * 
+   * @param element {HTMLElement}         Element to animate
+   * 
+   * @returns {Promise}                   Resolved when the animation is done
+   */
   enter(element) {
     return Promise.resolve(false);
   }
 
+  /**
+   * Execute a 'leave' animation on an element
+   * 
+   * @param element {HTMLElement}         Element to animate
+   * 
+   * @returns {Promise}                   Resolved when the animation is done
+   */
   leave(element) {
     return Promise.resolve(false);
   }
 
+  /**
+   * Add a class to an element to trigger an animation.
+   * 
+   * @param element {HTMLElement}         Element to animate
+   * @param className {String}            Properties to animate or name of the effect to use
+   * 
+   * @returns {Promise}                   Resolved when the animation is done
+   */
   removeClass(element, className) {
     return Promise.resolve(false);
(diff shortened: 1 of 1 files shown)


zewa666:
@EisenbergEffect I've discussed those changes before hand with @Aaike and will also apply the possible changes to the CSS-Animator. As from my side, this one is ok to merge :+1:

EisenbergEffect:
Has this been based solely on velocity? or have we confirmed that this abstraction would also work with something like greensock, for example?

Aaike:
I was going to create a GSAP implementation next ;)

Velocity has the same syntax as jQuery's `$.animate`.
`animate(element,propertiesOrEffectName,options)`

But GSAP is slightly different : `tweenlite.to(element,duration,options)`

However i think velocity's signature allows for more flexibility
The duration is separate in GSAP but this can be extracted from the `options` object.

The idea playing effects by name is done through "labels" in their TimeLine implementation.
The implementation for playing different "effects" would have to be done by creating a timeline for each effect , or to register an effect in one shared timeline.
This value can be the retrieved from the 2nd argument from the velocity/jquery signature.

I had something like this in mind for implementing GSAP with the new functions:
```
class GreensockAnimator{

  effects = {};

  function animate(element,effectName,options){
    if(typeof effectName == "string"){
      if(this.effects[effectName]){
        this.effects[effectName].play();
      }
    }else{
      TweenLite.to(element,options.duration,options)
    }
  }

  function registerEffect(effectName,properties){
    var tl = new TimelineLite();
    tl.to(element,options.duration,options);
    this.effects[effectName] = timeline;
  }
  
  function unregisterEffect(effectName,properties){
    delete this.effects[effectName];
  }
  
}
```

Other than GSAP i would also like to implement [TinyAnimate](https://github.com/branneman/TinyAnimate)
that one has a very different signature but it can be made to work with the same base functions.

The reason i want to implement all three is to let users choose between the perfect balance of features and filesize : 

- GSAP TweenLite (26 kb)
- GSAP TweenMax (104 kb)
- Velocity (35kb)
- TinyAnimate (3.8kb)

EisenbergEffect:
Excellent. Yes, we wanted to have the abstraction for exactly this reason. I can merge this in, since it sounds like you've looked over the various apis. We can fix things up as  you implement the different animators. I don't think it will break anyone since it's only animator implementors that will have to deal with the changes at this point.

PR #103, opened 2015-06-18
fix(BindableProperty): use strict equality operators

### LINKED ISSUE #100: show.bind does not handling falsy values inside a repeat
link to plunkr that shows the issue : http://plnkr.co/edit/DyiuUD?p=preview

title should have been show.bind does not handle falsy values :P

jdanyow:
Fixes #100

### DIFF
--- a/src/bindable-property.js
+++ b/src/bindable-property.js
@@ -199,7 +199,7 @@ class BehaviorPropertyObserver {
   setValue(newValue){
     var oldValue = this.currentValue;
 
-    if(oldValue != newValue){
+    if(oldValue !== newValue){
       if(this.publishing && this.notqueued){
         this.notqueued = false;
         this.taskQueue.queueMicroTask(this);
@@ -218,7 +218,7 @@ class BehaviorPropertyObserver {
 
     this.notqueued = true;
 
-    if(newValue != oldValue){
+    if(newValue !== oldValue){
       if(this.selfSubscriber !== null){
         this.selfSubscriber(newValue, oldValue);
       }


jdanyow:
this change fixes this scenario:
```javascript
export class App {
  model = [
    {test:null,title:"1"},
    {test:null,title:"2"},
    {test:null,title:"3"},
    {test:null,title:"4"},
    {test:null,title:"5"},
    {test:null,title:"6"}];
}
```
```html
<div repeat.for="item of model">
  <div show.bind="item.test">${item.title}</div>
</div>
```

We may need to change the `show` behavior to handle scenarios where `test` is undefined instead of null.

EisenbergEffect:
Cool. Thanks for taking this on. On a related note, I'm thinking we may want to enable value converters to be used by bindable properties. This would enable behavior properties, like show/if, to automatically convert their incoming data to the proper type. We could even use TS metadata to set up default converters based on type, and do some nice things for ES6 with a fluent api. 

For TS, we could write:

```
export class MyCustomElement {
  @bindable foo:boolean = false;
}
```

Assuming they turned on TS metadata, we could interpret that as this:

```
bindable({name:'foo', defaultValue: false, converter:BooleanConverter});
```

For ES6 we could enable something like this:

```
export class MyCustomElement {
  @bindable.boolean foo = false;
}
```

Thoughts?

jdanyow:
Not sure, I'm having trouble thinking of a use case for bindable property types other than boolean.  With a boolean bindable property we could use a converter to coerce truthy/falsy values to true and false.  If you have a number or Date bindable property and want to bind it to a string you need to know the format.  Same is true for the opposite scenario.  

Would there be a way to avoid having the components make assumptions about formats?  

There's probably a bunch of other scenarios I'm not thinking of.

Issue #111, opened 2015-06-24
add extra animation events for the `animate` method


Aaike:
some events should be added for the `animate` method of the base animator.

- `animation:animate:begin`
- `animation:animate:active`
- `animation:animate:done`

zewa666:
as discussed with you @Aaike I've added the sequence events, along the missing timeout events for add/removeClass as well. All the other discussed events can be added later as we need them

EisenbergEffect:
It's in now :)

PR #112, opened 2015-06-25
feat(anim): additional animation events

zewa666:
adds additional animation events, also incorporating new interface methods

fixes issue https://github.com/aurelia/templating/issues/111


Aaike:
looks good !

PR #115, opened 2015-06-28
fix(ViewCompiler): handle html attribute interpolation consistently

jdanyow:
(work in progress)

### DIFF
--- a/src/view-compiler.js
+++ b/src/view-compiler.js
@@ -47,6 +47,7 @@ export class ViewCompiler {
   static inject() { return [BindingLanguage]; }
   constructor(bindingLanguage){
     this.bindingLanguage = bindingLanguage;
+    this.InterpolationBindingExpression = bindingLanguage.inspectAttribute({ valueConverterLookupFunction: () => null }, 'foo', '${foo}').expression.constructor;
   }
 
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
This isn't quite right. I'm going to close this out. In fact we will have to revert your previous change in the compiler as well. I'm sorry I didn't catch the problem. The issue is that you cannot remove attributes at that point during compilation. The reason is that you might be iterating attributes related to "lifted content" ie an if or repeat node. If you remove the attribute, then it will be gone when that node gets compiled for the if/repeater thus causing the compiles template to be incorrect. If attributes are going to be removed, they must be removed inside the if block below: https://github.com/jdanyow/templating/blob/cbff88687c4328b78865d755f4f95652218bf92c/src/view-compiler.js#L272

EisenbergEffect:
Actually, you would want to do it directly before this line: https://github.com/jdanyow/templating/blob/cbff88687c4328b78865d755f4f95652218bf92c/src/view-compiler.js#L281 That's where the au-target class gets added. So, you would want to loop over the expressions and remove the corresponding attributes. The binding language may need to add an additional property to indicate an removed attribute. I'm not sure off the top of my head.

PR #122, opened 2015-07-04
Adds support for .ts views.

### LINKED ISSUE #121: TypeScript view-model support.
https://github.com/aurelia/templating/blob/master/src/view-strategy.js#L75-L79

If you are using SystemJS 0.18.0 which has runtime transpilation for TypeScript files, then the moduleId will resolve to `*.ts`.  When looking for the view, `.js` suffix is not found so `view.ts.html` is loaded (which likely doesn't exist).

Zoltu:
Fixes #121

### DIFF
--- a/src/view-strategy.js
+++ b/src/view-strategy.js
@@ -73,7 +73,7 @@ export class ConventionalViewStrategy extends ViewStrategy {
   }
 
   static convertModuleIdToViewUrl(moduleId){
-    var id = moduleId.endsWith('.js') ? moduleId.substring(0, moduleId.length - 3) : moduleId;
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@Zoltu Thanks! If you would like, there's another place. In framework/plugins there's some code that would need the same fix. Interested in submitting another PR for that as well?

Zoltu:
I don't mind, though I'm not sure where to find it.  I searched for `html` in templating repo to find this one, same search in `framework` doesn't yield promising results.

Zoltu:
Ah, found it.  I'll fix.

PR #145, opened 2015-08-05
fix(templating): Use correct import for core-js

davismj:
We were previously using `import core from core-js` which generates d.ts that do not match the corejs d.ts. This is now updated to `import * as core 'core-js'`, which resolves TypeScript compilation warnings.

closes aurelia/framework#177

### DIFF
--- a/src/bindable-property.js
+++ b/src/bindable-property.js
@@ -1,4 +1,4 @@
-import core from 'core-js';
+import * as core from 'core-js';
 import {hyphenate} from './util';
 import {bindingMode} from 'aurelia-binding';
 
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
@davismj Was there some reason that we could use `import from 'core-js'`?

davismj:
`import from 'core-js'` is invalid. `import 'core-js'` is broken in babel-dts-generator. Additionally, [the dts](https://github.com/borisyankov/DefinitelyTyped/blob/master/core-js/core-js.d.ts) is consistent with this particular import format.

PR #151, opened 2015-08-17
d.ts fix - optional property defaultBindingMode

milunka:
For d.ts generation i have made defaultBindingMode property in customAttribute decorator as optional.


EisenbergEffect:
Thanks!

PR #186, opened 2015-10-07
chore(tests): initialize PAL

jdanyow:
@EisenbergEffect not sure if this is right- it's what I had to do to get the tests passing


EisenbergEffect:
Not quite right. You need to install pal-browser as a dev dependency instead. You can do that with `jspm install aurelia-pal --dev` You don't want it listed in the main dependencies section but in the devDependencies section.

jdanyow:
fixed this up

PR #187, opened 2015-10-09
fix(build): ViewEngine depends on ModuleAnalyzer

jdanyow:


EisenbergEffect:
Was this causing an issue? If not, I would like to leave it as is...I think the dependencies are actually circular.

EisenbergEffect:
Hmm. Think I might have run into this. Will check it out...

PR #216, opened 2015-11-18
fix(child, children): Determine actual target by key

Ixonal:
TypeScript seems to leave descriptor undefined, so using they key to determine what to actually get the metadata for. Using typeof instead of checking for existence because empty strings ('') are coerced to false, though that SHOULD never happen.


EisenbergEffect:
Thanks for the contribution!

PR #244, opened 2016-01-09
fix(controller): stop element from inheriting scope

jdanyow:


src/controller.js
```diff
@@ -120,14 +119,24 @@ export class Controller {
...
     }
 
+    let overrideContext;
     if (this.view !== null) {
       if (skipSelfSubscriber) {
         this.view.viewModelScope = scope;
       }
-
-      this.view.bind(this.viewModel, createOverrideContext(this.viewModel, scope.overrideContext));
+      if (this.viewModel === scope.overrideContext.bindingContext) {
```
jdanyow: This fixes a bug where an `overrideContext` was becoming it's own `parentOverrideContext`.

src/controller.js
```diff
@@ -120,14 +119,24 @@ export class Controller {
...
+        overrideContext = scope.overrideContext;
+      } else if (this.instruction.inheritBindingContext) {
+        overrideContext = createOverrideContext(this.viewModel, scope.overrideContext);
+      } else {
+        overrideContext = createOverrideContext(this.viewModel);
+        overrideContext.__parentOverrideContext = scope.overrideContext;
```
jdanyow: capture the parent scope, without making it available to AccessScope... we may need it later

src/controller.js
```diff
@@ -120,14 +119,24 @@ export class Controller {
...
+      }
+      this.view.bind(this.viewModel, overrideContext);
     } else if (skipSelfSubscriber) {
-      this.viewModel.bind(context, scope.overrideContext);
+      overrideContext = Object.assign({}, scope.overrideContext);
+      overrideContext.parentOverrideContext = scope.overrideContext.__parentOverrideContext;
```
jdanyow: clone the overrideContext so we can hook up the parent scope without impacting other views/controllers that are using this overrideContext instance

jdanyow:
@EisenbergEffect how does this look now?

EisenbergEffect:
I'm assigning to myself so I can review and merge this wee. Last week was a bit crazy due to travelling but I should be able to get to it in the next few days and get a release out.

PR #255, opened 2016-01-29
feat(element-events)

joelcoxokc:
Feature: Element Events

Class to be used along side Element inside the dependency container. 
Handles adding, removing, and dispatching element events


EisenbergEffect:
@joelcoxokc Take a look at the changes I made. I fixed the doc comments and indicated optional params. I also fixed lint errors.

PR #288, opened 2016-03-09
feat(ViewSlot): save override context during binding for later eval

mbroadst:
When you call `bind` on a ViewSlot, the bindingContext is stored on the slot itself so that optimizations can occur in future binds. This patch simply enforces the same behavior for the override context, so that similar optimizations can be made using that information.

**NOTE:** I ran into the need for this during implementation of a more performant grid implementation I'm working through with @jdanyow. Since each of my "views" in a particular set of rows (for an array view strategy) are, in fact, ViewSlots (each containing a view for the `<td>` cells), it would be incredibly useful for me to have this information in that context.


EisenbergEffect:
@jdanyow Can you review?

jdanyow:
good catch @mbroadst - thank you :beers:

PR #309, opened 2016-03-28
doc(custom element): add custom element basics doc

AshleyGrant:
Let's consider this a first draft of this documentation. 

Questions:
Are there any other topics that need to be covered in the "basics" documentation?
Should we add working examples to go along with the example code?
Is it okay that I'm planning on splitting custom element documentation out in to multiple files?
Why is "A Christmas Story" the greatest Christmas movie of all time, and how did I watch Top Gun while writing this, but come up with examples that didn't involve Maverick?

### DIFF
--- a/doc/article/en-US/templating-custom-elements.md
+++ b/doc/article/en-US/templating-custom-elements.md
@@ -1,0 +1,206 @@
+---
+{
+  "name": "Templating: Custom Elements Basics",
+  "culture": "en-US",
+  "description": "An overview of the Aurelia templating engine's custom element functionality basics. Custom Elements are used to add custom components to Aurelia applications.",
+  "engines" : { "aurelia-doc" : "^1.0.0" },
+  "author": {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
I think I'd rather have it all in one doc. We have a TOC UI that is built based on the section titles. We are improving that int he next version of the docs app to make it more visible.

EisenbergEffect:
This is a good start. A couple of notes before moving on with additional documentation:

* There's a typo where you say `customAttribute` on accident.
* I'd prefer to not use the Skeleton application when explaining things. We may change that in the future or it may go away entirely and I'd rather the docs not be dependent on that as an example.

Other than that, it looks pretty good. If you can address those two issues, we can merge this in and then you can pick up with adding more details.

AshleyGrant:
Alright, I fixed the typos where I used `customAttribute` and changed the wording to not mention the Skeleton project. I also changed the file name to not say basics.

PR #312, opened 2016-03-29
ensure correct cache lookup

stoffeastrom:


src/module-analyzer.js
```diff
@@ -214,7 +214,7 @@ export class ModuleAnalyzer {
     let vs;
     let resourceModule;
 
-    resourceModule = this.cache[moduleId];
+    resourceModule = this.cache[moduleId] && this.cache.hasOwnProperty(moduleId);
     if (resourceModule) {
       return resourceModule;
     }
```
jdanyow: this looks strange to me- would that return `true` instead of what's in the cache?

Should it be:
```js
if (this.cache.hasOwnProperty(moduleId)) {
  return this.cache[moduleId];
}
```

stoffeastrom: Oops. Yes that's right :)

stoffeastrom: @jdanyow fixed

src/module-analyzer.js
```diff
@@ -213,10 +213,9 @@ export class ModuleAnalyzer {
     let conventional;
     let vs;
     let resourceModule;
-
-    resourceModule = this.cache[moduleId];
-    if (resourceModule) {
-      return resourceModule;
+;
+    if (this.cache.hasOwnProperty(moduleId)) {
+      return this.cache[moduleId];
```
bryanrsmith: It might be simpler to create `this.cache` with `Object.create(null)`.

EisenbergEffect: I was wondering about that! I think I'd prefer that.

Issue #270, opened 2016-02-19
Feature request: support $this with enhance


JeroenVinke:
I noticed that we can use $this in Aurelia templates to get a reference to the context, but it returns `undefined` when `enhance` is doing the compilation.

plunker: http://plnkr.co/edit/ZI88Uu?p=preview

jdanyow:
Updated to latest Aurelia bits: https://gist.run/?id=7c8dc8af39bf26046e374abba409328f

@EisenbergEffect is this really a bug?  Is there any way for the templating engine to know what the binding context is in scenarios where a DOM element is passed to the `enhance` method instead of a `EnhanceInstruction` with a `bindingContext` property?

https://github.com/aurelia/templating/blob/master/src/templating-engine.js#L80

EisenbergEffect:
Correct, it's not a bug if a binding context isn't passed in.

JeroenVinke:
Passing the context to bindingContext allows $this to work, so that's great. But now it doesn't pick up the overrideContext: https://gist.run/?id=b9939a488d6158ebef4285eee1c1ebb8

Do I still need to call .bind() myself? How would we pass the overrideContext to the enhance call as [this line](https://github.com/aurelia/templating/blob/master/src/templating-engine.js#L94https://jdanyow.github.io/rjs-bundle/bundles/aurelia.js) only passes one argument to .bind() which is the bindingContext. this);
    view.attached();
```

EisenbergEffect:
Ah, well that's a bug. That must have been missed when we added the override context implementation. @JeroenVinke Any interest in submitting a PR to add overrideContext to the EnhanceInstruction as well as pass that through to the bind call?

JeroenVinke:
Sure, I'll take a stab at it.

JeroenVinke:
OK, so now I'm able to pass an overrideContext and enhance will pass that to bind(). But now the real issue is showing.

Check out this gist: https://gist.run/?id=cda6266d95db146fd58699cd3b27e043. Note that this gist doesn't use the latest aurelia-templating yet, so it doesn't pass a bindingContext or overrideContext to enhance(). Instead, a .bind() call is used on the View that's returned from enhance(). The issue is the same.

As you see in the gist I enhance two divs. With one I pass an overrideContext to bind() and with the other I don't. `$this` only works when you don't use an overrideContext. Strangely `$this.someVar` works both scenarios

JeroenVinke:
[...]

PR #323, opened 2016-04-20
fix(templating-engine): allow for overrideContext in enhance

JeroenVinke:
https://github.com/aurelia/templating/issues/270


jdanyow:
nice work @JeroenVinke :+1:  much appreciated.

PR #324, opened 2016-04-20
feat(ViewSlot): moving Views across the slot

niieani:
Needed for: https://github.com/aurelia/templating-resources/pull/215.
Fixes: https://github.com/aurelia/templating-resources/issues/213.

### DIFF
--- a/src/view-slot.js
+++ b/src/view-slot.js
@@ -163,6 +163,25 @@ export class ViewSlot {
       }
     }
   }
+  
+  /**
+   * Moves a view across the slot.
+   * @param sourceIndex The index the view is currently at.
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@jdanyow @martingust Related to the other perf work from @niieani Can you review?

jdanyow:
looks good to me other than the missing test for this- @niieani- hold on while we review the other PR and if that one looks good we'll just need test for both (should be easy, the repeat and the view-slot already have a test suite) and then we can merge.

niieani:
@jdanyow @martingust Added tests for the new `move` method. However it seems you have some tests which fail on templating-resources@master. Can either of you check?

PR #339, opened 2016-05-06
Html behavior basic inheritence

levan-g:
Allows to inherit bindable properties and childBindings

### DIFF
--- a/src/child-observation.js
+++ b/src/child-observation.js
@@ -37,7 +37,7 @@ export function child(selectorOrConfig: string | Object): any {
   return createChildObserverDecorator(selectorOrConfig, false);
 }
 
-class ChildObserver {
+export class ChildObserver {
   constructor(config) {
     this.name = config.name;
     this.changeHandler = config.changeHandler || this.name + 'Changed';
--- a/src/html-behavior.js
+++ b/src/html-behavior.js
@@ -1,4 +1,4 @@
-import {Origin} from 'aurelia-metadata';
+import {Origin, metadata} from 'aurelia-metadata';
 import {ObserverLocator, Binding} from 'aurelia-binding';
 import {TaskQueue} from 'aurelia-task-queue';
 import {Container} from 'aurelia-dependency-injection';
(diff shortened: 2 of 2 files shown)


EisenbergEffect:
In your estimation, what are the limitations of this implementation?

levan-g:
this implementation is limited to bindable properties and childBindings. it's hard to configure other stuff like targetShadowDOM etc...

EisenbergEffect:
I think we're going to close this for now. We may consider this again in the future. I think we'd like to encourage the community to use compositional techniques (such as decorators that apply behaviors to a class) over inheritance.

atsu85:
@EisenbergEffect, I guess there are cases, when this PR could be handy, despite Your suggestion

> use compositional techniques (such as decorators that apply behaviors to a class) over inheritance

As I understand, there is a problem when adding properties/functions to a class with decorator when using TypeScript. TypeScript compiler doesn't understand that decorator has added properties/functions and when You try to access that property/function, You have two options to satisfy the TS compiler type checker:
1) manually add the fields/functions to the class (with correct type/signature) so compiler also understand that these fields/functions exist and can be used in the code
2) cast the instance to `any` at every call site that needs to use the members added by the decorator

Additionally creating decorators takes much more effort and understanding compared to parent class for this case.

So at the moment I'm not sure if I should be happy or sad regarding the decision to close this PR ;)

EisenbergEffect:
It's not in scope for our v1 release. We can consider it again in the future, but I think we need to nail down a few more things first.

With respect to TypeScript, that's something you should really follow up on with them. That seems like a severe deficiency in their type system since decorators have a main use cases of doing exactly what I'm talking about here. I'd be interested to see what the TS team is planning with respect to that.

levan-g:
For now TypeScript supports merging interface with class
```ts
  interface SomeBehavior {
      bindableProp1: string;
      bindableProp2: string;
  }
  class SomeBehavior {
  }
```
We can use this to implement decorators that apply behaviors to a class

Basic idea would be: 

```ts
export class InkBehavior {
    
    @bindable color: string;

    @bindable opacity: string;

    @bindable disabled: boolean;

    /**
     * Lifecycle event that will be called from controller
     */
    attached() {
        ...
    }
}

@behaviors(InkBehavior)
export class MyCustomElement {

}

/**
 * This provides intellisense for TypeScript
 */
export interface MyCustomElement extends InkBehavior {}
```
If you like it I can work on PR

EisenbergEffect:
I'm not sure you'll be able to make it work exactly like that because decorators, like bindable, have to be applied to the correct target. You can give it a try though.

levan-g:
What about Dependency Injection?

EisenbergEffect:
For DI, you could grab the inject array or function and alter it.

PR #368, opened 2016-06-02
feat(module-analyzer): Enable registering ViewEngineHooks by convention.

davismj:
depends on https://github.com/aurelia/binding/pull/423.

### DIFF
--- a/src/module-analyzer.js
+++ b/src/module-analyzer.js
@@ -3,6 +3,7 @@ import {Container} from 'aurelia-dependency-injection';
 import {TemplateRegistryEntry} from 'aurelia-loader';
 import {ValueConverterResource} from 'aurelia-binding';
 import {BindingBehaviorResource} from 'aurelia-binding';
+import {ViewEngineHooksResource} from 'aurelia-binding';
 import {HtmlBehaviorResource} from './html-behavior';
 import {viewStrategy, TemplateRegistryViewStrategy} from './view-strategy';
 import {ViewResources} from './view-resources';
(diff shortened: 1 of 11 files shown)


src/module-analyzer.js
```diff
@@ -3,6 +3,7 @@ import {Container} from 'aurelia-dependency-injection';
 import {TemplateRegistryEntry} from 'aurelia-loader';
 import {ValueConverterResource} from 'aurelia-binding';
 import {BindingBehaviorResource} from 'aurelia-binding';
+import {ViewEngineHooksResource} from 'aurelia-binding';
```
alvarezmario: @davismj now that you are at it, why not make just one import from aurelia-binding?

src/module-analyzer.js
```diff
@@ -274,6 +275,9 @@ export class ModuleAnalyzer {
         } else if (conventional = BindingBehaviorResource.convention(key)) {
           resources.push(new ResourceDescription(key, exportedValue, conventional));
           metadata.define(metadata.resource, conventional, exportedValue);
+        } else if (conventional = ViewEngineHooksResource.convention(key)) {
```
alvarezmario: @davismj the following 2 lines are exactly as the above, so why not move this to the previous `else-if`?

alvarezmario: @davismj In fact, I think you can do a little refactor here starting on https://github.com/davismj/templating/blob/6cf33d84b7ded2d5b3cefcb9bc79b3f02c42e3d8/src/module-analyzer.js#L272. 

```
else if (conventional = 
  ValueConverterResource.convention(key) 
  || BindingBehaviorResource.convention(key) 
  || ViewEngineHooksResource.convention(key)
) {
```
this looks like a nice refactor, don't you agree?

src/module-analyzer.js
davismj: I thought about it, but I don't do a lot of contributing and I wanted to keep with the current conventions / style.

src/module-analyzer.js
davismj: Sort of the same reason as above. This also applies to the existing code for `ValueConverterResource` and `BindingBehaviorResource`, so I just followed the convention.

davismj:
Nice suggestions @alvarezmario, I'll update my PR.

src/module-analyzer.js
alvarezmario: Oh, I see, it makes sense, but they probably will like it :smile: ping @EisenbergEffect

EisenbergEffect: Feel free to collapse those into a single import. It's just an accidental oversight in this case.

PR #380, opened 2016-06-13
feat(ViewSlot animateView): Attach the animation strategy as a method…

joelcoxokc:
… on the viewSlot in order to allow reconfigurability

Add ViewSlot.prototype.animateView():Promise
The purpose of this Pull request is to attach the request for animations to the viewSlot class.
Attaching the request for animations will allow someone to override the way a view is animated in and out, by simply returning a promise from the animateView method.

### DIFF
--- a/src/view-slot.js
+++ b/src/view-slot.js
@@ -51,7 +51,7 @@ export class ViewSlot {
    *   @param  direction  The animation direction enter|leave
    *   @returns null or a Promise
    */
-  animateView(view:View, direction:String<enter|leave> = 'enter'):Promise<any> {
+  animateView(view:View, direction:String = 'enter'):Promise<any> {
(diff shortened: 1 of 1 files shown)


src/view-slot.js
```diff
@@ -46,6 +46,20 @@ export class ViewSlot {
   }
 
   /**
+   *   Runs the animator against the first animatable element found within the view's fragment
+   *   @param  view       The view to use when searching for the element
+   *   @param  direction  The animation direction enter|leave
+   *   @returns null or a Promise
+   */
+  animateView(view:View, direction:String<enter|leave> = 'enter'):Promise<any> {
```
EisenbergEffect: Joel, that isn't valid type syntax. It should just be `string`. If you can fix that, I think we can get this in today.

PR #381, opened 2016-06-13
feat(module-analyzer): Enable registering ViewEngineHooks by convention

davismj:
Includes the `ViewEngineHooksResource` class. Replaces https://github.com/aurelia/templating/pull/371.


EisenbergEffect:
@davismj I'm really sorry....but can close this and submit a new PR? ;) We just did a major re-work in the build process and there are conflicts. If you can just drop the same code in and then fix the file list in paths, that should do it.

Issue #379, opened 2016-06-11
Uncaught TypeError: Cannot read property 'splice' of undefined


JeroenVinke:
In order to try out the new shadow dom stuff I altered the aurelia-templating and aurelia-pal-browser dependencies to use github.

I am getting this error now:

> Uncaught TypeError: Cannot read property 'splice' of undefined

This happens here:
https://github.com/aurelia/templating/blob/master/src/child-observation.js#L249

I believe that the ` items`  declaration should happen outside the `if`

Stack trace:
```
onAdd @ aurelia-templating.js:4242 Uncaught TypeError: Cannot read property 'splice' of undefined
onChildChange @ aurelia-templating.js:4084
```

EisenbergEffect:
Your property is probably not set to an array. We just need to fix this to make it provide a default array if it's not set.

JeroenVinke:
you're right, I misread the code (the declaration is correct)

JeroenVinke:
@EisenbergEffect initializing the array fixed it for me. 

I am now noticing that the `@children` array is still empty in ` attached()`, and that [this line](https://github.com/aurelia/templating/blob/master/src/child-observation.js#L249) is called after `attached()` has taken place. Is this a timing issue (and do you want a skeleton-navigation repro) or is this expected?

EisenbergEffect:
You can create a repro, but I'm not sure we will be able to do much about it. I'll need to investigate.

JeroenVinke:
Turns out that this is caused by bluebird. Repro is [here](https://github.com/JeroenVinke/children-timing-issue).

When I disable bluebird the array is filled from `attached()`. We're OK with removing bluebird though

JeroenVinke:
I have no idea why, but this issue only occurs when the **minified** version of bluebird is used.

https://gist.run/?id=14ac85668b3ca27dcd8ad6d3f6579fb0

If you switch bluebird from minified to non-minified in index.html the @children array is ready from attached.

Ah well..

PR #387, opened 2016-06-15
fix(child-observation): initialize @children array before using it

JeroenVinke:
fixes https://github.com/aurelia/templating/issues/379


EisenbergEffect:
Much thanks!

PR #393, opened 2016-06-16
feat(view-engine-hooks-resource): Add ViewEngineHooksResource

davismj:
Allows for generating ViewEngineHooks resources via convention or decorator.

Replaces https://github.com/aurelia/templating/pull/381.


EisenbergEffect:
Got it in. Thanks!

PR #394, opened 2016-06-18
doc(custom-element): add slot documentation

AshleyGrant:


doc/article/en-US/templating-custom-elements.md
```diff
@@ -272,15 +269,289 @@ Aurelia custom elements utilize the "slot based" content projection standard fro
...
+          </slot>
+        </div>
+      </div>
+    </slot-to-mixed-slot>
+  </source-code>
+</code-listing>
+
+## [Slot Implementation Limitations](aurelia-doc://section/9/version/1.0.0)
+
+At the time this document was written, no browser supports Shadow DOM v1 slots. Thus, Aurelia has implemented the specification. It also works with *template controllers* such as `if` and `repeat` which can dynamically generate content. We've fixed up our `@child` and `@children` decorators to understand the new model as well.
```
atsu85: @AshleyGrant
> It also works with *template controllers* such as `if` and `repeat` which can dynamically generate content

I think it worth rephrasing, so it would be very clear, that slots must not be nested inside template controllers, but slot content can have template controllers?

Note, I've added [tests for it to aurelia-template-lint](https://github.com/MeirionHughes/aurelia-template-lint/pull/27/files#diff-fa0e0cc88c52bc15a5a1ff71151bf248R47) to check the violation of this limitation at build-time ;)

Issue #418, opened 2016-07-11
ViewLocator type error when using Webpack


heruan:
The `CompositionContext` interface has `viewModel` defined as:
```javascript
/**
* The view model url or instance for the component.
*/
viewModel?: any;
```
but if it's set to a view model instance when using Webpack, I get this error:
```
Uncaught (in promise) TypeError: Cannot read property 'endsWith' of undefined(…)
    convertOriginToViewUrl @ aurelia-templating.js:705
    ConventionalViewStrategy @ aurelia-templating.js:563
    createFallbackViewStrategy @ aurelia-templating.js:700
    getViewStrategy @ aurelia-templating.js:691
    (anonymous function) @ aurelia-templating.js:4367
```
because `viewModel` seems assumed to be a string.

To reproduce simply use this as `main.js` in skeleton-esnext-webpack:
```javascript
import {bootstrap} from 'aurelia-bootstrapper-webpack';
import {App} from "./app";

bootstrap((aurelia) => {
  aurelia.use.standardConfiguration().developmentLogging();
  aurelia.start().then(() => aurelia.setRoot( aurelia.container.get(App), document.body) );
});
```

This happens using Webpack only, the same logic in skeleton-esnext works as expected.

EisenbergEffect:
I believe this is a limitation in Webpack due to the fact that it doesn't function as a true module loader.

@niieani Can you look into this please? This issue has come up repeatedly. Either there's a bug in the Webpack loader or this just ins't possible with Webpack. If it isn't possible, we need to include documentation.

@heruan As a workaround, you can always provide the module id. Another possibility would be to manually associate the module id with the view model like so

```javascript
import {Origin} from 'aurelia-metadata';

export class MyClass {
}

Origin.set(MyClass, new Origin('./my-class', 'MyClass'));
```

If you need, you could create a decorator to do this as well, potentially leveraging the module id constants that Webpack makes available. I believe it's `module.id` but I'm not sure. @niieani can confirm. So, you could do something like this:

```javascript
function origin(id, name) {
  return function(target) {
   Origin.set(target.constructor, new Origin(id, name || 'Default'));
  }
}

@origin(module.id)
export class MyClass {

}
```

heruan:
Thank you @EisenbergEffect, I hit this also when mapping a route in the router with `moduleId` set to an external module such as "module-name": using Webpack it fails with `Error: Cannot find module './module-name'`, whereas using JSPM it loads the first export in "module-name" as expected.

[...]

PR #425, opened 2016-07-16
fix(view-locator): check 'origin.moduleId' instead of 'origin'

niieani:
`if (origin)` was always evaluating to true, because when no Origin is found, an [`unknownOrigin` object is returned](https://github.com/aurelia/metadata/blob/7f50c4cc55ef6c6bf00ae3ffe14d0f97a14ce39e/src/origin.js#L4).
Instead, checking for 'origin.moduleId' will revitalize the currently-dead ifs.
This should mean better and earlier error handling for situations like: https://github.com/aurelia/templating/issues/418

### DIFF
--- a/src/view-locator.js
+++ b/src/view-locator.js
@@ -31,7 +31,7 @@ export class ViewLocator {
 
       viewStrategy.assert(value);
 
-      if (origin) {
+      if (origin.moduleId) {
         value.makeRelativeTo(origin.moduleId);
       }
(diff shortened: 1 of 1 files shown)


niieani:
Also note that we currently don't have any tests for unset/improper `Origin`'s, which is the reason this bug went under the radar. Adding such a test would be a great way to start for any new team-members or community contributors.

PR #434, opened 2016-07-24
feat(view-engine): add a decorator for declaring requires on the ViewModel

niieani:
This is a non-breaking, low-impact change, as it is only applied when the specified metadata value is set. 
Might seem like there were many changes, but in fact it's mostly adding the same, optional parameter `target` to all those classes that implement `loadViewFactory`.

This basically allows declaring `<require>`s in the ViewModel instead of the View (or in addition to it). My use case was to be able to plug-in some `ViewEngineHooks` via a decorator.

Example usage:

```ts
@viewResources('./awesome-view-engine-hooks')
export class MyViewModel {
  // ...
}
```

equivalent of:

```html
<template>
  <require from="./awesome-view-engine-hooks"></require>
  <!-- ... -->
</template>
```

The decorator can take in multiple parameters, they can also be objects like `{ src, as }` or actual external Classes (in which case its Origin's moduleId is added).


EisenbergEffect:
Interestingly, I just had a use case for this where I needed to load css for a component that had no view. I'll take a look tomorrow. I'm a bit nervous about this with the release this week though.

EisenbergEffect:
@niieani Did you test this? I don't see where the target is ever being passed to the view strategy.

niieani:
I did test it locally, yes, but you are correct - I forgot to commit a change to one more file where the target is being called. Added that commit now. 
I know we're close to the release, I promise this is the last thing; only fixes and docs from now on. 📜 
Not sure if it will solve your use-case as-is, because with `NoViewStrategy`, the underlying `loadViewFactory` is [never called](https://github.com/niieani/templating/blob/17472d43a8ee6ed2cd403c190b7bb469cacee4c7/src/view-strategy.js#L127). I can try to think how to get this feature in though, shouldn't be too hard.

niieani:
@EisenbergEffect I've solved your use case and added some basic docs about the new functionality. You can now also add dependencies directly on the `noView` decorator, like this:

```ts
@noView('./some-dependency', './another-dep.css')
```

niieani:
@EisenbergEffect note that the tests didn't fail, just that the CircleCI has timed out **after** they completed successfully.

PR #437, opened 2016-07-26
feat(interfaces.js): add Component lifecycle interfaces

StrahilKazlachev:
add TS convenience interfaces as discussed in aurelia/router#375


build/paths.js
```diff
@@ -25,6 +25,7 @@ var paths = {
 };
 
 paths.files = [
+  'interfaces.js',
```
StrahilKazlachev: @niieani could you check if adding `interfaces.js` here is appropriate?

src/interfaces.js
```diff
@@ -0,0 +1,51 @@
+import {View} from './view';
```
StrahilKazlachev: @atsu85, @niieani could you check the interfaces

build/paths.js
niieani: yeah, that looks fine.

src/interfaces.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+interface ComponentUnbind {
+    /**
+    * Implement this hook if you want to perform custom logic after the component is detached and unbound.
+    */
+    unbind(): void;
+}
+
+interface ComponentGetViewStrategy {
```
EisenbergEffect: This doesn't work for all components. It only works for composed components and routed components.

build/paths.js
EisenbergEffect: You can't add it at the top because ViewStrategy is not yet defined I think. I'll try to patch it up.

src/interfaces.js
```diff
@@ -0,0 +1,51 @@
...
+import {ViewStrategy} from './view-strategy';
+
+interface ComponentCreated {
+    /**
+    * Implement this hook if you want to perform custom logic after the constructor has been called.
+    * At this point in time, the view has also been created and both the view-model and the view 
+    * are connected to their controller. The hook will recieve the instance of the "owningView". 
+    * This is the view that the component is declared inside of. If the component itself has a view,
+    * this will be passed second.
+    */
```
atsu85: It might be worth documenting, that when a view is cached its created callback is only called once on the initial creation, not when it is reused.

PR #455, opened 2016-08-25
refactor(view-{compiler,factory}): use Element.prototype.classList

Gheoan:


EisenbergEffect:
I would need to have @jdanyow perf-test this. I'm a little worried about whether it would degrade a critical code path or not. It's looks nice though :)

Gheoan:
Definetly. Another competitor would be `element.className`.

jdanyow:
@Gheoan here's a gist you can use to compare setAttribute, classList and className: https://gist.run/?id=72ba5c3588d6375fee052fc8358a4606

Gheoan:
Here is a comparison: https://gist.run/?id=c6f0af8bb63c695cf0ca2b3eacdab6e4&sha=040748f3bf785efc1e2a6714bd5a51dc832215a5.

I believe repaint performance should be tested too.

EisenbergEffect:
@jdanyow Where are we at with this? Do we need to make changes or keep the code the same?

EisenbergEffect:
It seems like classList is the fastest.

jdanyow:
here's what I got for results (lenovo thinkpad x1 carbon touch):

Chrome
```
running...
setAttribute x 481,968 ops/sec ±1.39% (61 runs sampled)
classList x 406,667 ops/sec ±0.34% (57 runs sampled)
className x 517,682 ops/sec ±0.58% (66 runs sampled)
Fastest is: className
```

Firefox
```
running...
setAttribute x 151,167 ops/sec ±3.36% (56 runs sampled)
classList x 166,265 ops/sec ±0.53% (66 runs sampled)
className x 173,773 ops/sec ±0.77% (64 runs sampled)
Fastest is: className
```

@Gheoan / @EisenbergEffect what did you guys see for results?

EisenbergEffect:
Chrome, Windows on MacBook Pro

running...
setAttribute x 518,528 ops/sec ±2.79% (64 runs sampled)
classList x 573,305 ops/sec ±1.21% (65 runs sampled)
className x 599,219 ops/sec ±4.57% (44 runs sampled)
Fastest is: className

EisenbergEffect:
I think in my comment above, I meant to say `className`. I'll check again on OSX a bit later.

jdanyow:
re-tested using `"C:\Program Files (x86)\Google\Chrome\Application\chrome.exe" --user-data-dir="C:\chrome\dev-sessions\perf" --enable-precise-memory-info --enable-benchmarking --js-flags="--expose-gc"`

```
running...
setAttribute x 472,026 ops/sec ±1.23% (88 runs sampled)
classList x 411,906 ops/sec ±0.42% (93 runs sampled)
className x 532,113 ops/sec ±0.48% (90 runs sampled)
Fastest is: className
```

Gheoan:
Chrome

running...
setAttribute x 144,812 ops/sec ±0.37% (61 runs sampled)
classList x 149,181 ops/sec ±0.87% (64 runs sampled)
className x 152,250 ops/sec ±0.50% (61 runs sampled)
Fastest is: className

Firefox

running...
setAttribute x 74,137 ops/sec ±0.88% (61 runs sampled)
classList x 75,124 ops/sec ±0.96% (62 runs sampled)
className x 78,946 ops/sec ±1.22% (60 runs sampled)
Fastest is: className

`className` seems faster, however I am still in favor of `classList`. Using `classList` prevents same class from being added multiple times and checking if the element already has a class is not needed.

Gheoan:
Also, it would be interesting to see how paint is affected in a benchmark such as dbmonster.

EisenbergEffect:
@Gheoan I don't think paint will be affected since this change affects the compile phase which happens only once.

@jdanyow What do you think? Should I make the change to using `className'?

jdanyow:
This PR (classList) results in nicer code. None of the options really stand out in terms of perf. I'm in board with accepting this PR or switching to className.

EisenbergEffect:
I might just switch it to class name. I'd like to ponder it a bit more.

EisenbergEffect:
Closing this for now. I think if we make a changes we should probably go with className since it's technically the fastest.

Issue #363, opened 2016-05-29
if.bind renders DOM element and subsequently removes the element after binding


nCubed:
Is the expected behavior of `if.bind`:
1. Render the HTML of the template
2. Evaluate the binding expression
3. Remove the element from the DOM if the binding expression is false

For example, I would like to only present an image if the binding property of the image has a value:

``` html
<div repeat.for="m of items">
    <img if.bind="m.image" src="http://example.com/images/${m.image}" />
    <div>${m.title}</div>
</div>
```

I've also tried to wrap the image within a div container:

``` html
<div repeat.for="m of items">
    <div if.bind="m.image"><img src="http://example.com/images/${m.image}" /></div>
    <div>${m.title}</div>
</div>
```

In both cases, the same behavior occurs where the `<img />` element is being output to the browser resulting in a _400 Bad Request_ of `http://example.com/images/`. And after the `<img />` is rendered, the element is subsequently removed from the DOM when `if.bind` returns false.

EisenbergEffect:
This definitely should not be the behavior. @martingust @jdanyow There was a recent change made in `if`. Is it possible that it caused this issue?

nCubed:
I pulled down the latest getting started es 2106 kit and can not reproduce this issue. I've been following the pluralsight tutorial; perhaps I've set something up incorrectly. 

For now, I'd say this issue should be on-hold. I'm going to take the es 2016 getting started kit and see if I can reproduce the same issue I encountered with the pluralsight tutorial.

EisenbergEffect:
The getting started kit may have some slightly older versions of dependencies from the latest patch releases, so it's possible the issue still exists.

nCubed:
I've pulled down the latest and have been able to reproduce the issue. From what I can tell, the issue only occurs when modifying the binding context. I've attached a working sample for your review.

1. Install the missing packages
1. Run the app
1. Go to the issue363 tab
1. Use the next button to page to the 2nd page of results

Notice the initial page will have 1 missing image for "Lea to the Rescue", but not a 400 bad request. Upon going to the next/previous pages, all missing images will have a 400 bad request.

Perhaps I've set something up incorrectly with the paged results?

[issue-363.zip](https://github.com/aurelia/templating/files/289950/issue-363.zip)

EisenbergEffect:
@jdanyow Can you or @martingust review this?

nCubed:
If this helps, I've re-created a much simpler example:

``` javascript
constructor() {
[...]

PR #466, opened 2016-09-23
fix(BehaviorPropertyObserver): publish immediately when TaskQueue already flushing

jdanyow:
fixes aurelia/binding#496, aurelia/binding#479, #363

Requires aurelia/task-queue#16


jdanyow:
@EisenbergEffect I've tested this change (and the corresponding task-queue change) in the following:

* the skeleton-navigation project
* the templating-resources test-suite
* each of the gists attached to the issues mentioned above ^^^

I think this is good to go. Let me know what you think.

PR #474, opened 2016-10-06
fix(view-engine): load template resources in order specified in view

AshleyGrant:
I'm not sure if this is exactly how we should accomplish this, but this code will force view resources to be loaded sequentially.

This could slow down loading views that have a bunch of resources, but we don't have much choice (other than more intense refactoring) if we want to guarantee resources are loaded sequentially.

Fixes aurelia/bootstrapper#47, aurelia/skeleton-navigation#659

### DIFF
--- a/src/view-engine.js
+++ b/src/view-engine.js
@@ -248,11 +248,14 @@ export class ViewEngine {
 
       //cause compile/load of any associated views second
       //as a result all globals have access to all other globals during compilation
-      for (i = 0, ii = allAnalysis.length; i < ii; ++i) {
-        allAnalysis[i] = allAnalysis[i].load(container, loadContext);
-      }
-
-      return Promise.all(allAnalysis).then(() => resources);
+      // using reduce instead of Promise.all to ensure resources are loaded in order
+      // they are included on the page
+      return allAnalysis.reduce((prev, curr) => {
+        return prev.then(() => {
+          return curr.load(container, loadContext);
+        });
+      }, Promise.resolve())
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@jdanyow Did you have some further thoughts on this? I'm a bit nervous about changing it.

AshleyGrant:
I'm not personally in favor of this change, and I wrote it. That being said, this is definitely something we need to fix, as currently, CSS resources are loaded in to the DOM in the order their requests complete, which is decidedly not ideal.

@jdanyow any chance you have the time on your hands to do this the "right" way?

jdanyow:
These were the thoughts I had:

https://gitter.im/aurelia?at=57f6dddcd6251fd1269aa9e1
https://gitter.im/aurelia?at=57f6e49b68f560d80ce3d4f4

Was thinking we'd track the order a style resource was loaded so we could use the sequence when inserting the style elements in the dom.

I think this would solve the problem but the guidance should probably be "essential style resources like bootstrap" should be bundled and loaded up-front, with a `<link>` tag. Optional style resources that need to be loaded on an as-needed basis could use `<require>`.

In other words, hitting this issue probably means there is some optimization work you could be doing to improve how your styles are loaded.

Does this make sense to you guys?

EisenbergEffect:
Yes. Before we do any work on this though I want to look at the CSS
resource code to see if there is a simpler solution by somehow fixing that.

EisenbergEffect:
Closing this since we're not going to go with this implementation.

PR #476, opened 2016-10-08
Fix incorrect parameter name in doc comment

ErikSchierboom:
The `resources` parameter was renamed in [this commit](https://github.com/aurelia/templating/commit/b4316233ff2b6365ae1796ba6c4e816e6d98d1b3), but the accompanying parameter comment was left unchanged.


EisenbergEffect:
Thanks!

PR #491, opened 2016-11-13
fix(view-compiler): fallback to node.data when wholeText is not implemented

niieani:
`node.wholeText` does not exist in `jsdom`'s DOM implementation (used e.g. for server-side rendering)

### DIFF
--- a/src/view-compiler.js
+++ b/src/view-compiler.js
@@ -158,7 +158,7 @@ export class ViewCompiler {
       return this._compileElement(node, resources, instructions, parentNode, parentInjectorId, targetLightDOM);
     case 3: //text node
       //use wholeText to retrieve the textContent of all adjacent text nodes.
-      let expression = resources.getBindingLanguage(this.bindingLanguage).inspectTextContent(resources, node.wholeText);
+      let expression = resources.getBindingLanguage(this.bindingLanguage).inspectTextContent(resources, node.wholeText || node.data);
       if (expression) {
         let marker = DOM.createElement('au-marker');
         let auTargetID = makeIntoInstructionTarget(marker);


src/view-compiler.js
```diff
@@ -158,7 +158,7 @@ export class ViewCompiler {
       return this._compileElement(node, resources, instructions, parentNode, parentInjectorId, targetLightDOM);
     case 3: //text node
       //use wholeText to retrieve the textContent of all adjacent text nodes.
-      let expression = resources.getBindingLanguage(this.bindingLanguage).inspectTextContent(resources, node.wholeText);
+      let expression = resources.getBindingLanguage(this.bindingLanguage).inspectTextContent(resources, node.wholeText || node.data);
```
niieani: I don't see a way to determine what text accessor property is available without introducing unnecessary boilerplate or performance penalty (e.g. with `hasOwnProperty` on the `node`). 

The safer way could be: `node.wholeText !== undefined ? node.wholeText : node.data`.

This is the issue on jsdom having it unimplemented: https://github.com/tmpvar/jsdom/issues/1397

jdanyow: There's a few options- we could do a one-time check in the templating package. We could expose this in the PAL. We could polyfill the feature in jsdom. We could PR the fix for jsdom.

Definitely would be more comfortable with your `node.wholeText !== undefined ? node.wholeText : node.data` approach.

jdanyow: I sent the PR: https://github.com/tmpvar/jsdom/pull/1652

I was thinking you could do something like this in templating/view-compiler.js while we wait for the jsdom release:

```js
import {DOM} from 'aurelia-pal';

const textAccessorProperty = DOM.createTextNode('test').wholeText === undefined ? 'textContent' : 'wholeText';

export class ViewCompiler {
   ....
```
then `...inspectTextContent(resources, node.wholeText)` would change to `...inspectTextContent(resources, node[textAccessorProperty])`

EisenbergEffect: @jdanyow How much work is involved in getting the PR to jsdom? We could do this in the mean time, but I'm a bit worried about how it might affect performance. Could we polyfill the property instead? Literally, monkey patch it.

jdanyow: I'm working on the PR now, there's a problem with the test for it which I should have fixed in a few mins. We definitely could monkey-patch jsdom. I don't think there's a perf issue with my approach, I've added some comments above to try and clarify what I meant.

Gheoan:
As a workaround in `pal-nodejs`'s dependencies, `jsdom` can be changed to @jdanyow fork `github:jdanyow/jsdom#wholeText`.

niieani:
I feel we should do what @Gheoan suggests since we cannot be sure how soon jsdom will review @jdanyow's PR. Otherwise we could be blocked indefinitely. @EisenbergEffect I'll do the dependency swap in pal-nodejs if that's okay with you, unless you'd prefer money-patching the original `jsdom` for now?

EisenbergEffect:
Let's go with the dependency swap, assuming @jdanyow is good with keeping that fork alive/updated until his PR is merged.

jdanyow:
sure- @niieani, let me know if you need me to do anything.

niieani:
So I just tried, it's not enough to install the github:jdanyow/jsdom#wholeText fork because it needs some generated files which aren't there in the repo. @jdanyow would need to publish his fork on NPM after running the jsdom build script, or make a #wholeTextBuilt branch which we could then use.

niieani:
OK, maybe a polyfill for jsdom is the best idea after all. Maintaining a fork is just too involved.
I've made a PR to [add the polyfill](https://github.com/aurelia/pal-nodejs/pull/19) to pal-nodejs.

EisenbergEffect:
Closing this since we merge the polyfill into our pal-nodejs.

PR #506, opened 2016-12-14
feat(SwapOrder) Concurrent enter/leave transitions

mroseboom:
Is related to feature request [aurelia/templating-resources#264](https://github.com/aurelia/templating-resources/issues/264)

### DIFF
--- a/src/composition-engine.js
+++ b/src/composition-engine.js
@@ -83,34 +83,32 @@ export class CompositionEngine {
     this.viewLocator = viewLocator;
   }
 
-  _createControllerAndSwap(context) {
-    function swap(controller) {
-      let swapStrategy = SwapStrategies[context.swapOrder] || SwapStrategies.after;
-      let previousViews = context.viewSlot.children.slice();
+  _swap(context, view) {
+    let swapStrategy = SwapStrategies[context.swapOrder] || SwapStrategies.after;
+    let previousViews = context.viewSlot.children.slice();
 
-      return swapStrategy(context.viewSlot, previousViews, () => {
-        return Promise.resolve(context.viewSlot.add(controller.view)).then(() => {
-          if (context.currentController) {
-            context.currentController.unbind();
-          }
-        });
-      }).then(() => {
-        if (context.compositionTransactionNotifier) {
-          context.compositionTransactionNotifier.done();
+    return swapStrategy(context.viewSlot, previousViews, () => {
+      return Promise.resolve(context.viewSlot.add(view)).then(() => {
+        if (context.currentController) {
+          context.currentController.unbind();
         }
-
-        return controller;
       });
-    }
+    }).then(() => {
+      if (context.compositionTransactionNotifier) {
+        context.compositionTransactionNotifier.done();
+      }
+    });
+  }
 
+  _createControllerAndSwap(context) {
     return this.createController(context).then(controller => {
       controller.automate(context.overrideContext, context.owningView);
 
       if (context.compositionTransactionOwnershipToken) {
-        return context.compositionTransactionOwnershipToken.waitForCompositionComplete().then(() => swap(controller));
+        return context.compositionTransactionOwnershipToken.waitForCompositionComplete().then(() => this._swap(context, controller.view));
       }
 
-      return swap(controller);
+      return this._swap(context, controller.view).then(() => controller);
     });
   }
 
@@ -204,30 +202,11 @@ export class CompositionEngine {
         let result = viewFactory.create(context.childContainer);
         result.bind(context.bindingContext, context.overrideContext);
 
-        let work = () => {
-          let swapStrategy = SwapStrategies[context.swapOrder] || SwapStrategies.after;
-          let previousViews = context.viewSlot.children.slice();
-
-          return swapStrategy(context.viewSlot, previousViews, () => {
-            return Promise.resolve(context.viewSlot.add(controller.result)).then(() => {
-              if (context.currentController) {
-                context.currentController.unbind();
-              }
-            });
-          }).then(() => {
-            if (context.compositionTransactionNotifier) {
-              context.compositionTransactionNotifier.done();
-            }
-
-            return result;
-          });
-        };
-
         if (context.compositionTransactionOwnershipToken) {
-          return context.compositionTransactionOwnershipToken.waitForCompositionComplete().then(work);
+          return context.compositionTransactionOwnershipToken.waitForCompositionComplete().then(() => this._swap(context, result));
         }
 
-        return work();
+        return this._swap(context, result).then(() => result);
       });
     } else if (context.viewSlot) {
       context.viewSlot.removeAll();
--- a/src/composition-engine.js
+++ b/src/composition-engine.js
@@ -86,21 +86,20 @@ export class CompositionEngine {
   _createControllerAndSwap(context) {
     function swap(controller) {
       let swapStrategy = SwapStrategies[context.swapOrder] || SwapStrategies.after;
+      let previousViews = context.viewSlot.children.slice();
 
-      swapStrategy(context.viewSlot, context.viewSlot.children[0], () => {
-        return Promise.resolve().then(() => {
-          return context.viewSlot.add(controller.view);
-        }).then(() => {
+      return swapStrategy(context.viewSlot, previousViews, () => {
+        return Promise.resolve(context.viewSlot.add(controller.view)).then(() => {
           if (context.currentController) {
             context.currentController.unbind();
           }
-
-          if (context.compositionTransactionNotifier) {
-            context.compositionTransactionNotifier.done();
-          }
-
-          return controller;
         });
+      }).then(() => {
+        if (context.compositionTransactionNotifier) {
+          context.compositionTransactionNotifier.done();
+        }
+
+        return controller;
       });
     }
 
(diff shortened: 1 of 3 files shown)


src/composition-engine.js
```diff
@@ -69,6 +69,37 @@ function tryActivateViewModel(context) {
   return context.viewModel.activate(context.model) || Promise.resolve();
 }
 
+class SwapStrategies {
```
jods4: This is copy-pasted from `templating-router`. We should probably do something about it, like import from `templating` in `templating-router`?

src/composition-engine.js
```diff
@@ -69,6 +69,37 @@ function tryActivateViewModel(context) {
...
+    return promise;
+  }
+
+  // animate the next view in after the current view has been removed
+  after(viewSlot, previousView, callback) {
+    return Promise.resolve(viewSlot.removeAll(true)).then(callback);
+  }
+}
+
+const swapStrategies = new SwapStrategies();
```
jods4: Do we actually need a class? Seems overkill as this is just a collection of named methods. Is there ever going to be a second instance?

Also, per my previous comment about duplication in `templating-router`, I suppose we would need to `export const swapStrategies`?

src/composition-engine.js
```diff
@@ -202,14 +241,22 @@ export class CompositionEngine {
...
-            if (context.compositionTransactionNotifier) {
-              context.compositionTransactionNotifier.done();
-            }
-
-            return result;
+          let swapStrategy;
+
+          swapStrategy = context.swapOrder in swapStrategies
+                  ? swapStrategies[context.swapOrder]
+                  : swapStrategies.after;
```
jods4: can be streamlined to ` = swapStrategies[context.swapOrder] || swapStrategies.after`.
The `in` check is implicit since if it is not `in`, we'll get `undefined` out.

src/composition-engine.js
```diff
@@ -85,18 +116,26 @@ export class CompositionEngine {
...
-        }
-
-        context.viewSlot.add(controller.view);
-
-        if (context.compositionTransactionNotifier) {
-          context.compositionTransactionNotifier.done();
-        }
-
-        return controller;
+      let swapStrategy;
```
jods4: Is there a reason for the assignment two lines below?
The code style around is to `let` and assign on the same line.

src/composition-engine.js
```diff
@@ -85,18 +116,26 @@ export class CompositionEngine {
...
+
+      swapStrategy = context.swapOrder in swapStrategies
+              ? swapStrategies[context.swapOrder]
+              : swapStrategies.after;
+
+      swapStrategy(context.viewSlot, context.viewSlot.children[0], () => {
+        return Promise.resolve().then(() => {
```
jods4: Is there a reason to lose a turn with an empty `Promise.resolve()`? I don't see a reason to.

If yes it should be commented.
If not, you're probably better off with: 
`Promise.resolve(context.viewSlot.add(controller.view)).then()`.

src/composition-engine.js
```diff
@@ -69,6 +69,37 @@ function tryActivateViewModel(context) {
   return context.viewModel.activate(context.model) || Promise.resolve();
 }
 
+class SwapStrategies {
+  // animate the next view in before removing the current view;
+  before(viewSlot, previousView, callback) {
+    let promise = Promise.resolve(callback());
```
jods4: Is this public API surface? What is the contract exactly?

My understanding is that in the PR code those methods are used internally only and never escape this module.
In that case, `callback()` is guaranteed to return a `Promise` already...

src/composition-engine.js
mroseboom: No, this is also a copy-paste of templating-router :)

EisenbergEffect:
@jods4 Can you make another review of the updated code?

jods4:
Looks good to me.

I noticed that the `swap` and `work` functions are 99% copy-pasted. I wish we could dedup them.
There are two differences: 
- the final result, but that comes from outside so it can easily be changed to a chained `.then()`
- and `controller.result` vs `controller.view` in `viewSlot.add()`. What's the difference between those two?

mroseboom:
@jods4 My understanding is that `swap` will be called when you use a view-model and `work` will be called when you only use a view on the `compose` element. But perhaps @EisenbergEffect could give you a better answer.

src/composition-engine.js
```diff
@@ -202,9 +205,16 @@ export class CompositionEngine {
...
 
         let work = () => {
-          return Promise.resolve(context.viewSlot.removeAll(true)).then(() => {
-            context.viewSlot.add(result);
-
+          let swapStrategy = SwapStrategies[context.swapOrder] || SwapStrategies.after;
+          let previousViews = context.viewSlot.children.slice();
+
+          return swapStrategy(context.viewSlot, previousViews, () => {
+            return Promise.resolve(context.viewSlot.add(controller.result)).then(() => {
```
EisenbergEffect: I don't see where the `controller` variable is coming from...

mroseboom: You're are right! `controller.result` should be `result`. I will update this.

jods4:
Now that this `result` has been clarified, it's clear that we could dedup that `work`/`swap` code.
All it takes is the view in parameter (`controller.view` or `result`) and chain the final result with a `.then(() => controller)` (or `.then(() => result)`).

src/composition-engine.js
```diff
@@ -204,30 +202,11 @@ export class CompositionEngine {
...
-              context.compositionTransactionNotifier.done();
-            }
-
-            return result;
-          });
-        };
-
         if (context.compositionTransactionOwnershipToken) {
-          return context.compositionTransactionOwnershipToken.waitForCompositionComplete().then(work);
+          return context.compositionTransactionOwnershipToken.waitForCompositionComplete().then(() => this._swap(context, result));
```
jods4: you don't return `result` after the swap here.

mroseboom: Indeed, good one.

zewa666:
Great contribution @mroseboom, looking forward to trying it out

EisenbergEffect:
@mroseboom Couldn't the router now use this existing implementation so that the code doesn't have to be duplicated?

mroseboom:
@EisenbergEffect Yes, that should be possible now. I'll make a pull-request within the templating-router repository for this.

mroseboom:
A little update, i've got the new swapStrategy working with the `router-view` component. 

Because i had some trouble to get this work in combination with the `removeMany` method and get the following error [aurelia/router#458](https://github.com/aurelia/router/issues/458). I've tried another implementation for the `swapStrategies` (see example below). Herefore i've added the [Array.isArray()](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_compatibility) method. Only now the second parameter `previous` could be an array or a single view.

@jods4 Please let me know which implementation you prefer, the current one or the one below where the second parameter `previous` can be both a array or single view?

After this i can make the pull-request within the `templating-router` branche.

**swapStragies**
```javascript
  // animate the next view in before removing the current view;
  before(viewSlot, previous, callback) {
    if (previous !== undefined) {
      return callback()
        .then(() => viewSlot[Array.isArray(previous) ? 'removeMany' : 'remove'](previous, true));
    }

    return callback();
  },

  // animate the next view at the same time the current view is removed
  with(viewSlot, previous, callback) {
    if (previous !== undefined) {
      return Promise.all([
        viewSlot[Array.isArray(previous) ? 'removeMany' : 'remove'](previous, true), 
        callback()
      ]);
    }

    return callback();
  },
```
**router-view (above implementation)**
```javascript
swap(viewPortInstruction) {
    let layoutInstruction = viewPortInstruction.layoutInstruction;
    // single view
    let previousView = this.view;

    let work = () => {
      let swapStrategy = SwapStrategies[this.swapOrder] || SwapStrategies.after;
      let viewSlot = this.viewSlot;

      swapStrategy(viewSlot, previousView, () => {
        return Promise.resolve(viewSlot.add(this.view));
      }).then(() => {
        this._notify();
      });
    };
```

**router-view (current implementation)**
```javascript
swap(viewPortInstruction) {
    let layoutInstruction = viewPortInstruction.layoutInstruction;
    // This works with the current swapStrategy implementation
    let previousViews = this.viewSlot.children.slice();

    // This didn't worked with the current swapStrategy implementation
    // let previousViews = [this.view];

    let work = () => {
      let swapStrategy = SwapStrategies[this.swapOrder] || SwapStrategies.after;
      let viewSlot = this.viewSlot;

      swapStrategy(viewSlot, previousViews, () => {
        return Promise.resolve(viewSlot.add(this.view));
      }).then(() => {
        this._notify();
      });
    };
```

jods4:
I think I prefer the version above because it calls the same methods as of today.
The call on `this.viewSlot.children.slice()` makes me a little uneasy, especially since the more intuitive `[this.view]` doesn't work. @EisenbergEffect what do you think?

The only drawback is that it doesn't read very well.
What if we extract the removal into a function? Something like:
```js
function remove(viewSlot, previous) {
  return Array.isArray(previous) ?
    viewSlot.removeAll(previous, true) :
    viewSlot.remove(previous, true);
}

export const SwapStrategies {
  before(viewSlot, previous, callback) {
    return previous === undefined ?
      callback() :
      callback().then(() => remove(viewSlot, previous));
  }

  with(viewSlot, previous, callback) {
    return previous === undefined ?
      callback() :
      Promise.all([callback(), remove(viewSlot, previous)]);
  }
}
```

It's the same code but I think it reads better. 
Don't copy-paste, I typed inside github ;)

EisenbergEffect:
I'm in agreement with @jods4. Let's prefer the version that calls the standard ViewSlot APIs. That helps ensure that if we make an internal change there in the future, we don't break this. If we can go with that along with the refactoring that @jods4 mentions, I think we should be good.

PR #510, opened 2016-12-26
feat: Support for `default property` binding on custom attributes

dkent600:
See templating-binding feature request:  https://github.com/aurelia/templating-binding/issues/56

Requires templating-binding PR:  https://github.com/aurelia/templating-binding/pull/108

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "typescript.tsdk": "./node_modules/typescript/lib"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "typescript.tsdk": "./node_modules/typescript/lib"
+}
--- a/config.js
+++ b/config.js
@@ -20,7 +20,7 @@ System.config({
     "aurelia-templating-binding": "npm:aurelia-templating-binding@1.1.0",
     "babel": "npm:babel-core@5.8.38",
     "babel-runtime": "npm:babel-runtime@5.8.38",
-    "core-js": "npm:core-js@1.2.7",
+    "core-js": "npm:core-js@2.4.1",
     "github:jspm/nodelibs-assert@0.1.0": {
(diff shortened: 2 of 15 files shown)


EisenbergEffect:
I kind of like the idea of using `primaryProperty` so it's less likely to be confused with `defaultValue`. What do you think? @jdanyow

jdanyow:
Sounds good to me- @dkent600 would you mind renaming defaultProperty and defaultBindable to primaryProperty? Then I'll get this merged. Will need another PR for templating-binding because I merged your original already. Make sure you pull down the latest

CasiOo:
I have a couple questions to the changes made to `config.json`.

Why is `core-js` being taken from version `2.4.1` to `1.2.7`?
Why is `npm:aurelia-templating@1.1.4` suddenly added to `config.json` when it wasn't needed before?

dkent600:
@CasiOo 

`npm:aurelia-templating@1.1.4` was added to the project only for testing purposes.
`core-js@1.2.7` -- I don't know.  I wasn't willing to mess with the version numbers as JSPM left them.

CasiOo:
@dkent600 Ah, because `aurelia-templating` is a dependency of `aurelia-templating-binding`.

The core-js version change still worries me.

dkent600:
@CasiOo I just reverted the core-js version numbers, downloaded the resulting package, built and ran all the tests, successfully.  The change is committed to this PR.

dkent600:
@jdanyow The property changes are commited.

CasiOo:
After this feature gets merged, there should be opened an enhancement task to remove the cyclic dependencies. Having `aurelia-templating-binding` as a dependency for the tests results in a cyclic dependency tree, because `aurelia-templating` will end up with a dependency on itself (the npm version).

Surely we should be able to come up with a test suit where `aurelia-templating-binding` isn't needed.

dkent600:
If anyone wants to enlist me further in this, let me know.

PR #517, opened 2017-01-15
feat(templating-engine): Register View with container on enhance

### LINKED ISSUE #516: Inject associated View instance when available
**I'm submitting a feature request**

* **Library Version:**
1.2.0

**Current behavior:**
When `@inject(View)`, a new `View` is created with no arguments. This produces an error.

If injected in an instance that has an associated `View`, the associated `View` should be injected.

**What is the motivation / use case for changing the behavior?**
This will allow an instance to handle its own lifecycle events.

davismj:
It took some verbose code to try to access a View from within an enhanced component. This change registers the View instance on the relevant container when calling enhance.

closes #516

### DIFF
--- a/src/templating-engine.js
+++ b/src/templating-engine.js
@@ -3,6 +3,7 @@ import {DOM} from 'aurelia-pal';
 import {Controller} from './controller';
 import {ModuleAnalyzer} from './module-analyzer';
 import {Animator} from './animator';
+import {View} from './view';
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@davismj Was there some reason you couldn't use the created callback to get the view instance?

EisenbergEffect:
@davismj Ping

PR #522, opened 2017-01-24
fix(view-slot): check for undefined child

zewa666:
in removeAll, check for undefined children before executing returnToCache

fixes issue https://github.com/aurelia/animator-css/issues/57

### DIFF
--- a/src/view-slot.js
+++ b/src/view-slot.js
@@ -339,7 +339,9 @@ export class ViewSlot {
 
       if (returnToCache) {
         for (i = 0; i < ii; ++i) {
-          children[i].returnToCache();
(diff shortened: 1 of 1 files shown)


src/view-slot.js
```diff
@@ -339,7 +339,9 @@ export class ViewSlot {
 
       if (returnToCache) {
         for (i = 0; i < ii; ++i) {
-          children[i].returnToCache();
```
EisenbergEffect: Let's eliminate the double lookup by storing the child in a local variable so we don't have to index it again if it is defined.

zewa666: oki doki makes sense

PR #535, opened 2017-03-14
feat(aliases): aliases for custom attributes

zewa666:
Allows to create aliases for custom attributes by providing an array of aliases to the customAttribute decorator, or by applying then on the targets static property aliases

### DIFF
--- a/src/decorators.js
+++ b/src/decorators.js
@@ -57,8 +57,12 @@ export function customElement(name: string): any {
 * @param name The name of the custom attribute.
 * @param defaultBindingMode The default binding mode to use when the attribute is bound wtih .bind.
 */
-export function customAttribute(name: string, defaultBindingMode?: number): any {
+export function customAttribute(name: string, defaultBindingMode?: number, aliases?: string[]): any {
   return function(target) {
+    if (aliases) {
+      target.aliases = aliases;
+    }
+
     let r = metadata.getOrCreateOwn(metadata.resource, HtmlBehaviorResource, target);
     r.attributeName = validateBehaviorName(name, 'custom attribute');
     r.attributeDefaultBindingMode = defaultBindingMode;
--- a/src/decorators.js
+++ b/src/decorators.js
@@ -55,7 +55,8 @@ export function customElement(name: string): any {
 /**
 * Decorator: Indicates that the decorated class is a custom attribute.
 * @param name The name of the custom attribute.
-* @param defaultBindingMode The default binding mode to use when the attribute is bound wtih .bind.
+* @param defaultBindingMode The default binding mode to use when the attribute is bound with .bind.
+* @param aliases The array of aliases to associate to the custom attribute.
 */
 export function customAttribute(name: string, defaultBindingMode?: number, aliases?: string[]): any {
   return function(target) {
--- a/src/decorators.js
+++ b/src/decorators.js
@@ -60,13 +60,10 @@ export function customElement(name: string): any {
 */
 export function customAttribute(name: string, defaultBindingMode?: number, aliases?: string[]): any {
   return function(target) {
-    if (aliases) {
-      target.aliases = aliases;
-    }
-
     let r = metadata.getOrCreateOwn(metadata.resource, HtmlBehaviorResource, target);
     r.attributeName = validateBehaviorName(name, 'custom attribute');
     r.attributeDefaultBindingMode = defaultBindingMode;
+    r.aliases = aliases;
   };
 }
 
--- a/src/html-behavior.js
+++ b/src/html-behavior.js
@@ -168,6 +168,12 @@ export class HtmlBehaviorResource {
   register(registry: ViewResources, name?: string): void {
     if (this.attributeName !== null) {
       registry.registerAttribute(name || this.attributeName, this, this.attributeName);
+
+      if (Array.isArray(this.target.aliases)) {
+        this.target.aliases.forEach( (alias) => {
+          registry.registerAttribute(alias, this, this.attributeName);
+        });
+      }
     }
 
     if (this.elementName !== null) {
--- a/src/html-behavior.js
+++ b/src/html-behavior.js
@@ -170,9 +170,16 @@ export class HtmlBehaviorResource {
       registry.registerAttribute(name || this.attributeName, this, this.attributeName);
 
       if (Array.isArray(this.target.aliases)) {
-        this.target.aliases.forEach( (alias) => {
-          registry.registerAttribute(alias, this, this.attributeName);
-        });
+        this.target
+          .aliases
+          .filter((a, pos, arr) =>
+             a !== this.attributeName &&
+             typeof a === 'string' &&
+             a.length > 0 &&
+             arr.indexOf(a) === pos)
+          .forEach( (alias) => {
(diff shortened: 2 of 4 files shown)


src/html-behavior.js
```diff
@@ -168,6 +168,19 @@ export class HtmlBehaviorResource {
...
       registry.registerAttribute(name || this.attributeName, this, this.attributeName);
+
+      if (Array.isArray(this.target.aliases)) {
+        this.target
+          .aliases
+          .filter((a, pos, arr) =>
+             a !== this.attributeName &&
+             typeof a === 'string' &&
+             a.length > 0 &&
+             arr.indexOf(a) === pos)
```
stsje: What is this last check good for? `arr.indexOf(a) === pos`

src/decorators.js
```diff
@@ -55,10 +55,15 @@ export function customElement(name: string): any {
...
 * Decorator: Indicates that the decorated class is a custom attribute.
 * @param name The name of the custom attribute.
-* @param defaultBindingMode The default binding mode to use when the attribute is bound wtih .bind.
+* @param defaultBindingMode The default binding mode to use when the attribute is bound with .bind.
+* @param aliases The array of aliases to associate to the custom attribute.
 */
-export function customAttribute(name: string, defaultBindingMode?: number): any {
+export function customAttribute(name: string, defaultBindingMode?: number, aliases?: string[]): any {
   return function(target) {
+    if (aliases) {
```
stsje: Couldn't we just always assign `target.aliases`, do we even need the check to see if it's undefined?

src/html-behavior.js
stsje: It seems that `registerAttribute` already checked for empty attribute name and if there's a duplicate.
I think the filter part can/should be removed all together

https://github.com/aurelia/templating/blob/master/src/view-resources.js#L6

stsje:
I think this `alias` functionality belongs in `view-resources`, just like the dependency injection container supports aliases. In my opinion the `customAttribute` shouldn't try and implement any aliasing functionality itself.

The implementation seems to be working good though!

AStoker:
@stsje, where do you imagine this functionality inside the `view-resources`?

stsje:
@AStoker I suggest expanding `ViewResources.registerAttribute` method with another parameter.
`registerAttribute(attribute: string, behavior: HtmlBehaviorResource, knownAttribute: string, aliases?: string[]): void`

Then just move what you have here into ViewResources instead. That's what I would have done at least.

AStoker:
@stsje, so straight up on the first register (https://github.com/aurelia/templating/pull/535/files#diff-68f429b0b6798ade8b97b54751efb74bR170)? I could see that. You would still need a way to hook into that functionality though, some way of telling the system that when you register this "whatever" that you have aliases to go along with it. So I still think you'd need some kind of modification to the customAttribute binding.

stsje:
Yes the decorator will probably still need the `aliases` parameter, but I would implement attribute aliasing in `view-resources`. I hope we understand each other :)

AStoker:
I think I'm tracking. It does seem to move the logic closer to where it should be (based off my limited understanding of the templating engine). @zewa666, what do you think?

src/html-behavior.js
zewa666: Good point, didn't look closer at registerAttribute

src/decorators.js
zewa666: Undefined wouldn't be an issue but what's with null or false? Actually it might make sense to do a proper array check here

zewa666:
I personally don't care where it actually is implemented, I just followed directions from @EisenbergEffect since I'm pretty new to templating as well

zewa666:
@stsje could you show us what you're thinking of where you'd move the part? From what I read its not like putting it in the registerAttribute part directly, so here https://github.com/aurelia/templating/blob/master/src/view-resources.js#L219.

What was your idea instead?

src/decorators.js
```diff
@@ -55,10 +55,15 @@ export function customElement(name: string): any {
...
+      target.aliases = aliases;
```
EisenbergEffect: Is there some reason that you are assigning this to the target and not to the metadata object below? The metadata object is where all the configuration for the custom attribute is.

zewa666: the reason is the use case for i18n, where the developer needs a chance to modify the aliases from outside. As such its easy to import the TCustomAttribute and overload the static property. It would be quite cumbersome to get access to the `metadata`, being honest right now I wouldn't even know how to do that.

AStoker: @EisenbergEffect, any update on this? Would be great to get this in to solve a few bugs (https://github.com/aurelia/i18n/issues/192, among other bugs that I have in my system)

EisenbergEffect: We should put that aliases on the behavior. That can be done with the following API:

```javascript
import {HtmlBehaviorResource} from './html-behavior';
import {metadata} from 'aurelia-metadata';

let r = metadata.getOrCreateOwn(metadata.resource, HtmlBehaviorResource, target);
r.aliases = ...
```

@zewa666 Can you make that change?

EisenbergEffect: Inject is a broad capability, not dependent on the html behavior system. The aliases are strictly linked to the behavior system, so I'd like to store that information along with the rest of the metadata.

zewa666: @EisenbergEffect I do understand your point but that does not solve the use case I was describing. If somebody consumes the i18n library, he needs a way to define the aliases for the TCustomAttribute. The consumer has no idea about HtmlBehaviorResources nor metadata nor anything.

Even requiring him to do

```
import {TCustomAttribute} from 'aurelia-i18n'

configure()... {
  TCustomAttribute.aliases = [...]
}
```

isn't the nicest api but its yet understandable. So with the above described metadata approach, how would the consumer of i18n define aliases for the TCustomAttribute?

zewa666: The only reason why I compared it to inject, is that it is simple to access and modify. If we get something as simple as a static property I'm totally happy. Going the decorator way isn't the solution though, as that is only for the creator of the Custom Attribute, but not the consumer, whereas the most interesting use case for Aliases is actually the Consumer being able to kinda *rename* the attribute to fit it's own dev pattern.

EisenbergEffect: Why not just add a helper method to do it?

```javascript
class TCustomAttribute {
  static configureAliases(aliases) {
    let r = metadata.getOrCreateOwn(metadata.resource, HtmlBehaviorResource, TCustomAttribute);
    r.aliases = aliases;
  }
}
```

AStoker: @EisenbergEffect, moved the saving of the aliases to the metadata itself. Tested it out with that helper method and it works just fine. Was this what you were thinking of?

EisenbergEffect:
@AStoker Yes, that's the idea.

AStoker:
@zewa666, this look good over here?

zewa666:
Alright we need to get the unit tests straight and remove the unnecessary checks as per @stsje. I'm on phone so can't fix it yet but can take a look at it hopefully tomorrow.

AStoker:
@EisenbergEffect @zewa666, should be good for a final review. Removed the extra filter as per @stsje's comment, and updated tests.

PR #537, opened 2017-03-19
fix(primaryProperty): infinite recursion with template controller

jods4:
Using `primaryProperty` on a `templateController` attribute results in infinite recursion.

The reason is that `instruction.originalAttrName` sole purpose is to be removed from the template but because `primaryProperty` override `attrName` with the option name, it doesn't remove the attribute from the template anymore, hence the loop.

I fixed this specific bug only.

**There are other usages of `attrName` in `view-compiler` and I find _highly suspicious_ that we should replace the real custom attribute name with one of its options property. This wouldn't happen when not using `primaryProperty` so smells very wrong to me. Someone with more templating knowledge than me should review these uses.**

If I'm correct about this being unwanted, another fix could be to not re-assign `attrName` rather than introducing an extra `originalAttrName` variable like I did.

CC: @dkent600 introduced this change.

### DIFF
--- a/src/view-compiler.js
+++ b/src/view-compiler.js
@@ -293,6 +293,7 @@ export class ViewCompiler {
     let attr;
     let attrName;
     let attrValue;
+    let originalAttrName;
     let instruction;
     let info;
     let property;
@@ -319,7 +320,7 @@ export class ViewCompiler {
 
     for (i = 0, ii = attributes.length; i < ii; ++i) {
       attr = attributes[i];
-      attrName = attr.name;
+      originalAttrName = attrName = attr.name;
       attrValue = attr.value;
       info = bindingLanguage.inspectAttribute(resources, tagName, attrName, attrValue);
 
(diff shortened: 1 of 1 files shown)


dkent600:
Cool!  Is it  possible you could to add a test for this case?

jods4:
@dkent600 we'll have one when aurelia/templating-resources#280 is merged because it depends on this to work.

Can you review if my change is the right thing to do or if you shouldn't modify `attrName` in the first place? I only patched the `templateController` behavior because I was not sure of the impact of other uses.

dkent600:
@jods4 Sure I'll take a look.  But caveat emptor:  Templating is not my area of expertise (I wasn't aware of templateControllers until now).  But I did mess around with this code a lot to implement the primaryProperty feature, so I won't be completely useless.

dkent600:
@jods4 Sorry it took me a while to look further into this!  I think this code change makes perfect sense.

Issue #96, opened 2015-06-15
Add a type option to bindable for type conversion


Aaike:
I was wondering if it would be a good idea to add an option to the `@bindable` decorator to make it convert a value to a different type than a string.

This can be useful when creating a custom element that accepts numeric or boolean options :

``` markup
<slider min="0" max="10" precision="0.1" snap="true"></slider>
```

Right now the values would have to be manually converted every time the variable is used.
perhaps an option could be added like this:

```
@bindable({type:"int"}) min;
@bindable({type:"int"}) max;
@bindable({type:"float"}) precision;
@bindable({type:"boolean"}) snap;
```

EisenbergEffect:
Absolutely. One of the ideas I have is to let you specify a ValueConverter as part of your bindable definition. This would then enable arbitrary type conversion to be built in to the property. To simplify common use cases, you could specify a type, and then we would select a built-in converter based on that type. In the case of TypeScript, we can even leverage the type metadata to do this automatically. Thoughts?

Aaike:
yes a value converter seems much more flexible :)

EisenbergEffect:
@Aaike I'm thinking of something like this:

For TS, we could write:

```
export class MyCustomElement {
  @bindable foo:boolean = false;
}
```

Assuming they turned on TS metadata, we could interpret that as this:

```
bindable({name:'foo', defaultValue: false, converter:BooleanConverter});
```

For ES6 we could enable something like this:

```
export class MyCustomElement {
  @bindable.boolean foo = false;
}
```

Thoughts?

Aaike:
Those defaults look great yes, and this also still allows custom converers for ES6 in the following form i assume ?
```
export class MyCustomElement {
  @bindable({converter:SomeConverter}) foo = false;
}
```

EisenbergEffect:
Yes. That's my thought.

brettveenstra:
If I was taking the lazy way and tried to bind a complex JSON object into my viewmodel, would I have to create a custom valueconverter or define a nested `@bindable` or define on the element itself?

jadrake75:
I think the datatype to the binding signature would be a great feature.  Just discovered our numeric inputs are converting to strings so this would have been nice to coerce the binding without having to use a value converter

jdanyow:
[...]

PR #558, opened 2017-07-31
feat(bindable): support coercing value

bigopon:
This PR add support for coercing value, before calling `setValue` on property, to address the issue: #96 

Usage is similar to @jdanyow 's proposal

```js
class MyViewModel {
  @bindable({ coerce: 'number' }) numProp
  @bindable({ coerce: 'boolean' }) boolProp
  @bindable({ coerce: val => convertValue(val) }) customCoerceProp
}
```

To modify / extends the built-in coerce types:

```js
import {coerces} from 'aurelia-templating';

coerces.string = function(a) {
  return a === null || a === undefined ? '' : a.toString();
}
```

To add more coerce types:

```js
coerces.longDate = function(a) {
  return new Date(a).toJSON();
}

// usage:
@bindable({ name: 'date', coerce: 'longDate' })
class DateControl {

}
```

All basic implementations:

```js

const numCons = Number;
const dateCons = Date;
const _isFinite = isFinite;
const _isNaN = isNaN;

export const coerces = {
  none(a) {
    return a;
  },
  number(a) {
    var val = numCons(a);
    return !_isNaN(val) && _isFinite(val) ? val : 0;
  },
  string(a) {
    return '' + a;
  },
  boolean(a) {
    return !!a;
  },
  date(a) {
    return new dateCons(a);
  }
};
```

To avoid breaking changes/ bug, any fail attempt to get coerce function will result into a identity function, which is `coerces.none`

cc @EisenbergEffect @jdanyow


bigopon:
@EisenbergEffect if this passes, it will be trivial to support your proposal:

```js
class MyViewModel {
  @bindable.number numProp
}
```

EisenbergEffect:
Cool. The other thing we'd want to do is support TypeScript metadata, so if they do this...

```
number: boolean;
```

...we should pick up the metadata and map it to a coerces function as well.

bigopon:
I have added the support for Typescript.

Now it supports this:

```ts

import {mapCoerceForClass} from 'aurelia-templating';

class Point {}

mapCoerceForClass(Point, 'point', function(value) {
  return new Point(value.split(' ').map(parseFloat));
});

class LineViewModel {
  point1: Point
  point2: Point
}
```

Basically you can register how to coerce a class. JavaScript users can also leverage this by setting metadata on property:

```js

import {mapCoerceForClass} from 'aurelia-templating';

class Point {}

mapCoerceForClass(Point, 'point', function(value) {
  return new Point(value.split(' ').map(parseFloat));
});

class LineViewModel {
  @Reflect.metadata('design:type')
  point1

    @Reflect.metadata('design:type')
  point2
}
```

@EisenbergEffect one thing I'm not sure & confident about is naming of `mapCoerceForClass` & `classCoerceMap` (this is a map holding Class definition and their string representation for look up in `coerces` export)

bigopon:
I just re-read the issue again and it seems you @EisenbergEffect  agreed with @niieani that it shouldn't use metadata to automatically coerce value. So I'm not actually sure about the changes related to the metadata.

EisenbergEffect:
Don't worry too much about the api yet. We can tweak that. If you'd like to keep working on this (awesome work btw!) then I'm wondering if you can investigate the binding library's `observable` decorator. I can imagine wanting to support this with both `observable` and `bindable`, so I'm wondering if we can put the code in the binding library somehow, or at least the core of it that is independent of the decorators.

src/behavior-property-observer.js
```diff
@@ -92,3 +117,51 @@ export class BehaviorPropertyObserver {
     this.removeSubscriber(context, callable);
   }
 }
+
+const numCons = Number;
+const dateCons = Date;
+const _isFinite = isFinite;
+const _isNaN = isNaN;
```
atsu85: @bigopon, could you explain the purpose of these 4 aliases - it looks like adding indirection that makes readability worse without obvious benefits

src/behavior-property-observer.js
```diff
@@ -92,3 +117,51 @@ export class BehaviorPropertyObserver {
...
+  boolean(a) {
+    return !!a;
+  },
+  date(a) {
+    return new dateCons(a);
+  }
+};
+
+/**@type {Map<Function, string>} */
+export const classCoerceMap = new Map([
```
atsu85: I think that instead of the two lines above You can write smth like:
`export const classCoerceMap: Map<Function, string> = new Map([`
or even better (more precise type):
`export const classCoerceMap: Map<(strValue: string) => any, string> = new Map([`
As it would be correct way to write type information for TypeScript variables/constants, and at the moment some Babel plugin is used to extract this information to generate API docs and definitions files for TS.
My only concern is that maybe the Babel plugin that extracts TS types doesn't understand so complex TS type expressions as i wrote in the second example

atsu85: This would also be beneficial when converting this repo to TS (as i understand, some Aurelia repos are already written in TypeScript and some are in middle of transitioning to TS)

src/behavior-property-observer.js
```diff
@@ -92,3 +117,51 @@ export class BehaviorPropertyObserver {
...
+  [Date, 'date']
+]);
+
+/**
+ * Map a class to a string for typescript property coerce
+ * @param Class {Function} the property class to register
+ * @param strType {string} the string that represents class in the lookup
+ * @param converter {function(val)} coerce function tobe registered with @param strType
+ */
+export function mapCoerceForClass(Class, strType, coerce) {
```
atsu85: perhaps type information could be moved next to parameters from function documentation - smth like
`export function mapCoerceForClass(Class: Function, strType: string, coerce: (strValue: string) => any) {`

atsu85:
@bigopon, nice work! I added couple of comments tough

src/behavior-property-observer.js
bigopon: It is expected that those function will be called with high frequency, so it is better to put them in the same scope with the calling function. This is to work around bundlers, which will add many scope layers containing maybe up to hundred of enumerable accessors.

src/behavior-property-observer.js
bigopon: You are right about this, could have inlined type def.
But the purpose of this variable is to hold a map between Classes and their coerce type (as string) in real look up table of `coerces` for later use. So the type would be `Map<Function, string>`

EisenbergEffect:
@jdanyow Can you work with @bigopon to address any issues and add support for this for both observable and bindable? This is a long time requested feature and one I know people will want to use.

src/behavior-property-observer.js
jdanyow: how about:
```
mapCoercionFunction(type: { new:(): any; }, typeName: string, coercionFunction: (value: string) => any)
```

src/decorators.js
```diff
@@ -87,30 +88,66 @@ export function templateController(target?): any {
...
+      /**
+       * Support for Typescript decorator, with metadata on property type.
+       * Will check for typing only when user didn't explicitly set coerce
+       */
+      if (!nameOrConfigOrTarget.coerce) {
+        propType = metadata.getOwn(metadata.propertyType, target, key2);
+        if (propType) {
+          nameOrConfigOrTarget = classCoerceMap.get(propType) || 'none';
+        }
+      }
```
jdanyow: Is this turning on coercion by default in any project that uses TypeScript with emit decorator metadata?

cc @EisenbergEffect

EisenbergEffect: I think we need to have a setting that allows people to opt into this. Turning it on automatically, at this point, would be a breaking change to people's apps.

bigopon:
@jdanyow  @EisenbergEffect 

I have:

  * fixed all naming issues
  * added a flag to turn off the type auto pickup by default. Also export a function to turn it on
  * added a function `createTypedBindable` to create bindable with fluent syntax `@bindable.string`.
  * created built in for `string`, `number`, `date`, `boolean` with `createTypedBindable`

Please let me know if I missed something.

src/behavior-property-observer.js
```diff
@@ -36,11 +60,12 @@ export class BehaviorPropertyObserver {
   * @param newValue The new value to set.
   */
   setValue(newValue: any): void {
-    let oldValue = this.currentValue;
+    const oldValue = this.currentValue;
+    const coercedValue = this.coerce !== undefined ? this.coerce(newValue) : newValue;
```
jdanyow: nit pick- positive case first please.  `this.coerce === undefined ? newValue : ....`

src/behavior-property-observer.js
```diff
@@ -13,15 +16,36 @@ export class BehaviorPropertyObserver {
...
+  }
+  setCoerce(coerce) {
+    let c;
+    switch (typeof coerce) {
+    case 'function':
+      c = coerce; break;
+    case 'string':
+      c = coerceFunctions[coerce]; break;
+    default: break;
+    }
```
jdanyow: nit pick- does the indentation of the switch block pass the lint rules?

src/behavior-property-observer.js
```diff
@@ -13,15 +16,36 @@ export class BehaviorPropertyObserver {
...
+    if (c === undefined) {
+      LogManager
```
jdanyow: could this be put in the `default:` switch case?  Then move the `this.coerce` assignment into the function and string cases, enabling removal of the `let c`?

src/behavior-property-observer.js
bigopon: I couldn't see it in `.eslint.json` in `aurelia-tools`. When I was doing this, I remembered I got error in linting for not indenting the `case`. And I tested with 1 more indent, it showed error. I think it's  following the rule.

src/behavior-property-observer.js
bigopon: 👍

bigopon:
@jdanyow

I have added typings for `bindable`, `BindableProperty`. `bindable.string`, ...etc.

I haven't added typings for `coerceFunctions` and `mapCoerceFunction` because it would be more desirable to have both `@observable` and `@bindable` to have same sort of functionality (per Rob). So I think we will need to move `coerceFunctions` and `mapCoerceFunction` there. There was a PR in `binding` for that but I messed up my history and it's gone. Can easily recreate though.

I don't know if this is the final desired behavior, as currently it still doesn't play nice together with two-way binding. Basically it can't be synced well with input, for current binding notification model. And it would take not trivial amount work to support it properly, consider syncing between a `date` bindable property to an `input`, or `select`. We can, however, encourage user to use `from-view` and typed bindable.

What do you think ?

bigopon:
@jdanyow @EisenbergEffect doc is ready too. oh docs ...

doc/article/en-US/templating-html-behaviors-introduction.md
```diff
@@ -297,7 +297,232 @@ The `@bindable` decorator isn't used when doing single value binding with a Cust
...
+    // Assume all incoming values are strings that look like this '5.1 6.2'
+    coerceFunctions.point = function(incomingValue) {
+      return incomingValue.split(' ').map(parseFloat);
+    };
+  </source-code>
+</code-listing>
+
+#### Usage with metadata for Typescript
+
+Typescript compiler has an option emit class fields with their types in metadata. `bindable` decorator can work with this via `usePropertyType` function, which is a property of `bindable` decorator:
```
atsu85: Perhaps instead of "an option emit" you meant "an option **to** emit"?

doc/article/en-US/templating-html-behaviors-introduction.md
```diff
@@ -297,7 +297,232 @@ The `@bindable` decorator isn't used when doing single value binding with a Cust
...
+  * `Number` to `'number'`
+  * `String` to `'string'`
+  * `Boolean` to `'boolean'`
+  * `Date` to `'date'`
+
+You can extend / modify this via export `coerceFunctionMap` of `aurelia-binding`:
+
+<code-listing heading="extend-coerce">
+  <source-code lang="TypeScript">
+    // import from 'aurelia-binding' if you are writing a plugin
```
atsu85: is it explained somewhere else, why plugins should import things from original package instead of `aurelia-framework` that exports things from other packages? I guess they both work, but 'aurelia-binding' should be preferred to avoid excessive dependencies (and to be tree-shaking friendly)?

doc/article/en-US/templating-html-behaviors-introduction.md
```diff
@@ -297,7 +297,232 @@ The `@bindable` decorator isn't used when doing single value binding with a Cust
...
+      point2
+    }
+  </source-code>
+</code-listing>
+
+In above example, `bindable` decorator auto picks up the type, set proper coerce function for `point1` and `point2`.
+
+#### Fluent syntax
+
+If you scroll back top a bit, you will notice we had fluent syntax decorator: `@bindable.number`. This, as described above, is a simplified and more expressive form of `coerce: 'number'`. Aurelia also provides a way to build your own fluent syntax `bindable` decorator. You can do this by:
```
atsu85: Perhaps instead of 
"If you scroll back **to**p a bit"
following would be better:
"If you scroll back **u**p a bit"

doc/article/en-US/templating-html-behaviors-introduction.md
```diff
@@ -297,7 +297,232 @@ The `@bindable` decorator isn't used when doing single value binding with a Cust
...
+    // This is to enable `@bindable.point`
+    createTypedBindable('point');
+
+    // Register a class to be used with coerce
+    coerceFunctionMap.set(Point, 'point');
+
+    // Add `point` coerce function
+    coerceFunctions.point = function(incomingValue) {
+      return incomingValue.split(' ').map(parseFloat);
+    };
```
atsu85: Perhapse there should be a helper to simplify registering fluent coerce syntax? For example instead of 
+ `import {coerceFunctions, coerceFunctionMap, createTypedBindable} from 'aurelia-framework';`
+ `createTypedBindable(..)`
+ `coerceFunctionMap.set(...)`
+ `coerceFunctions.... = function...` 
it could be done for example like this:
```TypeScript
import {coerceHelper} from 'aurelia-framework';
coerceHelper.addFuluentBindableExtension('point', Point, (incomingValue) => {
    return incomingValue.split(' ').map(parseFloat);
});
```

bigopon: Right, I was feeling this is too much call too, for API designed, I think @jdanyow @EisenbergEffect will need to give call on this.

@atsu85 Thanks for helping, I ran out of energy half way through the docs 😱 , many typos and unneeded stuff

EisenbergEffect:
@bigopon Can you give me an update on where we are at on this. Seems like it stalled out. I'd love to have types bindables/observables though.

thomas-darling:
This looks really awesome, @bigopon - great work, and much appreciated!

I do have a couple of concerns though:

* The use with TypeScript seem too complicated. If the property type is `number`, and the value being set is a `string`, it should just always coerce by default - having to explicitly turn that on for every single bindable property will be a huge pain, and it would make no sense, as you already specified that the desired type, and the compiler will complain if you try to treat it as anything else.

  Not coercing by default will just cause bugs, exactly as it does today, where the compiler and developer thinks a property will be of the type specified, but then at runtime, it ends up being something completely different. In our code, this is probably our single biggest source of bugs.

* As I understand it, boolean attributes such as `disabled` are not handled automatically, but requires a coerce function to be specified. I think handling boolean attributes according to the HTML spec should be the default behavior.

  Previously, we also talked about having a `reflect` option on the `bindable` decorator, which would control whether the property value is reflected to the DOM attribute or not - I think that’s also still relevant.

* The concept of coerce functions do not support two way binding, as far as I can tell, and I’m wondering whether we need this new concept at all? We already have value converters, which essentially do the same thing, and support two-way conversion - maybe we should just allow those to be used, instead of introducing a new concept?

  The only issue with that, would be how to support boolean attributes, where they have to not just convert the value, but also modify the attribute on the element. However, since boolean attributes are really a very special case, I think it might be ok to just hardcore support for that...

  So if that `reflect` option I mentioned is set to true, and `toView` returns true, then the attribute is automatically added with an empty string as value. If `toView` returns false, null or undefined, then the attribute is automatically removed. Maybe with an exception for attributes whose name starts with `aria-`, as those do not follow the standard behavior, but requires the values “yes” and “no”.

  It’s a simple solution, but it gets the job done, and it would also solve previously reported issues with attributes, such as `src` and `href`, which should only be present if they actually have a value.

  An alternative approach might be to, when the value converter is used for coercion and that `reflect` option is true, pass the element and attribute name to the `toView` and `fromView` functions of the converter. Then the converter could make the nessesary attribute changes - and again, if we choose this approach, then I think the boolean converter we provide out of the box should do this by default.

Now, we might need options to opt-out of some of this default behavior until app code is updated, but I think that will be the best long-term strategy, as opposed to littering the app code with options to opt-in on every bindable property.

Thoughts, @bigopon, @EisenbergEffect?

bigopon:
@thomas-darling 

> The use with TypeScript seem too complicated. If the property type is number ...

Are you referring to `bindable.usePropertyType(true)` ? If that is true, then it shouldn't be a concern as it's supposed to be called only once. It is there for compatibility reason.

> As I understand it, boolean attributes such as disabled are not handled automatically ...

View model properties are not supposed to be used with view only, but also interacted within JavaScript / TypeScript. so by default I don't think we can treat `@bindable.boolean` similar to Html `boolean` attribute. Consider `""`, it's `true` for Html `boolean` attributes but false for Js/Ts `boolean`. I think best we can do is to have another default: `@bindable.booleanAttr` where it treat `""`, or `"required", "readonly"` as `true`

> The concept of coerce functions do not support two way binding, as far as I can tell, and I’m wondering whether we need this new concept at all? ...

It could be nice if we can avoid it, but for me, value converter are somewhat view related logic, while typed `bindable`/ `observable` are supposed to work seamlessly in script side. Beside that, it would bring huge overhead to make current value converter work within Ts/Js I think. @EisenbergEffect & @jdanyow can give better feedback

> Previously, we also talked about having a reflect option on the bindable decorator ...

It's is being discussed here #560 . Ideally we should have it reviewed after this PR is decided to be merged or not

> The concept of coerce functions do not support two way binding ...

It depends on the target of the binding, if it's a custom element / custom attribute, then it works well and as expected, except some cases like gotchas in the [observable PR](https://github.com/aurelia/binding/pull/623). For Html built-ins, value are always coerced to string so it's pretty much nothing we can do.

atsu85:
@EisenbergEffect, @jdanyow - i've been really dreaming about this issue since the beginning. Contrary to what @thomas-darling said, this feature is useful and easy to use with TypeScript, as reflect metadata is available at runtime and the fact that we'd need to call `bindable.usePropertyType(true)` just once during Aurelia app configuration doesn't make absolutely any difference for me.

Fantastic work @bigopon, i hope it will be merged soon!

CasiOo:
I have some refactor suggestions, getting inspiration from the validation plugin.
In validation we're using static methods and properties to add custom rules without having an instance. Maybe we should be doing the same here?
`ValidationRules.customRule(...);`

So here are my suggestions
```
// Add fluent binding syntax for coercion converter by default.
// Optionally state a property type. Useful for TS users or JS using reflection metadata.
Coercion.customConverter(key, converter, propertyType?);

// Interpret of property types should be set on a Coercion object and not on bindable in my opinion.
Coercion.usePropertyType(true);
// or
Coercion.interpretPropertyType(true);
```

doc/article/en-US/templating-html-behaviors-introduction.md
```diff
@@ -297,7 +297,223 @@ The `@bindable` decorator isn't used when doing single value binding with a Cust
...
+  </source-code>
+</code-listing>
+
+As you have noticed, `coerce` property is used to tell the decorator to find the right coercion function to convert incoming value. It can also be a function if you prefer:
+
+<code-listing heading="number-field${context.language.fileExtension}">
+  <source-code lang="ES 2015">
+    export class NumberField {
+      @bindable label
+      @bindable({ coerce: val => Number(val) }) value
```
gheoan: This can be `@bindable({ coerce: Number }) value` or is that confusing?

bigopon: thanks I'll fix it the next commit 👍

Issue #283, opened 2016-03-02
Reflecting properties to attributes


EisenbergEffect:
@levan1993 commented on [Wed Mar 02 2016](https://github.com/aurelia/binding/issues/333)

https://www.polymer-project.org/1.0/docs/devguide/properties.html#attribute-reflection

Is there something similar in aurelia?

---

@EisenbergEffect commented on [Wed Mar 02 2016](https://github.com/aurelia/binding/issues/333#issuecomment-191352485)

At the moment, there is not, but I think we could add this in the future. I'll add it to my todo list and update that here.

glen-84:
@EisenbergEffect,

I'm using a jQuery plug-in, and it looks for a `selected` attribute on the selected option, but `selected.bind` only sets the property, and not the attribute -- would this case also be solved by attribute reflection, or would it be better to have something like `selected.bind-attr` (a new feature to support some of the cases discussed in #76)?

Alexander-Taran:
@EisenbergEffect added to vNext candidates list.
Can be closed

thomas-darling:
See my comment here: https://github.com/aurelia/binding/issues/347#issuecomment-451935018

Alexander-Taran:
@thomas-darling your comment there is about vNext.
I do not try to kill your requests. I referenced them specifically in vNext monorepo to be able to come back to them.
I'm trying to clean up the mess of 800+ open issues across 20+ repositories.
Thank you for understanding.

PR #560, opened 2017-08-31
feat(BindableProperty): Reflect prop to attribute

bigopon:
This PR addresses: aurelia/binding#347 @AshleyGrant 

Note:

* It doesn't support app root component
* It doesn't support compose element
* It still set the attribute on `containerless` element
* It only has setup test so far
* [Demo](https://gist.run/?id=995e1c6a5e1ccecdec46f8421f862b72)

**Usage with `bindable` decorator:

```js
class MyComponent {
  @bindable({
    reflect: true // This instruct it to map prop -> attr 1:1
  })
  prop1

  @bindable({
    // Alternatively can be passed a function as instruction how to reflect the property to the element
    reflect(element: Element, propertyName: string, newVal, oldVal) {
      element.setAttribute(propertyName, `${oldVal} -- ${newVal}`);
    }
  })
  prop2
}
```

cc @EisenbergEffect @jdanyow

### DIFF
--- a/src/bindable-property.js
+++ b/src/bindable-property.js
@@ -35,7 +35,13 @@ export class BindableProperty {
   * Creates an instance of BindableProperty.
   * @param nameOrConfig The name of the property or a cofiguration object.
   */
-  constructor(nameOrConfig) {
+  constructor(nameOrConfig: string | {
+    defaultBindingMode?: number,
+    reflect?: boolean | {(el: Element, newVal, oldVal): any},
+    name?: string,
+    attribute?: any,
+    changeHandler?: string
+  }) {
     if (typeof nameOrConfig === 'string') {
       this.name = nameOrConfig;
     } else {
--- a/src/bindable-property.js
+++ b/src/bindable-property.js
@@ -37,7 +37,7 @@ export class BindableProperty {
   */
   constructor(nameOrConfig: string | {
     defaultBindingMode?: number,
-    reflect?: boolean | {(el: Element, newVal, oldVal): any},
+    reflect?: boolean | {(el: Element, name: string, newVal, oldVal): any},
     name?: string,
     attribute?: any,
     changeHandler?: string
--- a/src/bindable-property.js
+++ b/src/bindable-property.js
@@ -35,7 +35,7 @@ export class BindableProperty {
   * Creates an instance of BindableProperty.
   * @param nameOrConfig The name of the property or a cofiguration object.
   */
-  constructor(nameOrConfig: string | Object) {
+  constructor(nameOrConfig) {
     if (typeof nameOrConfig === 'string') {
       this.name = nameOrConfig;
     } else {
@@ -61,6 +61,10 @@ export class BindableProperty {
     behavior.properties.push(this);
     behavior.attributes[this.attribute] = this;
     this.owner = behavior;
+
+    if (this.reflect) {
+      behavior.registerReflection(this.name, this.reflect);
+    }
 
     if (descriptor) {
       this.descriptor = descriptor;
--- a/src/bindable-property.js
+++ b/src/bindable-property.js
@@ -64,12 +64,14 @@ export class BindableProperty {
   * @param descriptor The property descriptor for this property.
   */
   registerWith(target: Function, behavior: HtmlBehaviorResource, descriptor?: Object): void {
+    let { reflect } = this;
     behavior.properties.push(this);
     behavior.attributes[this.attribute] = this;
     this.owner = behavior;
 
-    if (this.reflect) {
-      behavior.registerReflection(this.name, this.reflect);
+    if (reflect) {
+      behavior._registerReflection(this.name, typeof reflect === 'function' ? reflect : propToAttr);
+      this._configureReflection(target);
     }
 
     if (descriptor) {
@@ -78,6 +80,38 @@ export class BindableProperty {
     }
 
     return undefined;
+  }
+  
+  _configureReflection(target) {
+    if (target.__reflectionConfigured__) return;
+
+    let method = 'propertyChanged';
+    let onChanged = target.prototype[method];
+    let hasHandler = !!onChanged;
+    
+    let alteredHandler;
+    if (hasHandler) { // avoid ternary to make it consisten with the rest ?
+      alteredHandler = function propertyChanged(name, newVal, oldVal) {
+        onChanged.call(this, name, newVal, oldVal);
+        let { __element__, __reflections__ } = this.__observers__;
+        if (!__element__) return;
+        __reflections__[name].call(this, __element__, name, newVal, oldVal)
+      }
+    } else {
+      alteredHandler = function propertyChanged(name, newVal, oldVal) {
+        let { __element__, __reflections__ } = this.__observers__;
+        if (!__element__) return;
+        __reflections__[name].call(this, __element__, name, newVal, oldVal);
+      }
+    }
+  
+    if (!Reflect.defineProperty(target.prototype, method, {
+      configurable: true,
+      value: alteredHandler
+    })) {
+      throw new Error(`Cannot setup property reflection on <${this.name}/> for ${target.name}`);
+    };
+    target.__reflectionConfigured__ = true;
   }
 
   _configureDescriptor(descriptor: Object): Object {
@@ -256,5 +290,20 @@ export class BindableProperty {
 
     observer.publishing = true;
     observer.selfSubscriber = selfSubscriber;
+  }
+}
+
+/**
+ * @private
+ * Used for avoid creating mapping function multiple times
+ * @param {Element} element 
+ * @param {string} propertyName 
+ * @param {any} newValue 
+ */
+function propToAttr(element, propertyName, newValue) {
+  if (newValue == null) {
+    element.removeAttribute(propertyName);
+  } else {
+    element.setAttribute(propertyName, newValue);
   }
 }
--- a/src/html-behavior.js
+++ b/src/html-behavior.js
@@ -510,5 +510,9 @@ export class HtmlBehaviorResource {
  * @param {any} newValue 
  */
 function propToAttr(element, propertyName, newValue) {
-  element.setAttribute(propertyName, newValue);
+  if (newVal == null) {
+    element.removeAttribute(propertyName)
+  } else {
+    element.setAttribute(propertyName, newValue);
+  }
 }
--- a/src/html-behavior.js
+++ b/src/html-behavior.js
@@ -323,13 +323,16 @@ export class HtmlBehaviorResource {
 
     if (element !== null) {
       element.au = au = element.au || {};
-      this._setupReflections(element, this.target);
     }
 
     let viewModel = instruction.viewModel || container.get(this.target);
     let controller = new Controller(this, instruction, viewModel, container);
     let childBindings = this.childBindings;
     let viewFactory;
+
+    if (element !== null) {
+      this._setupReflections(element, viewModel);
+    }
 
     if (this.liftsContent) {
       //template controller
@@ -405,49 +408,28 @@ export class HtmlBehaviorResource {
   /**
    * Allow a bindable property on custom element to register how to reflect prop value to attribute
    * @param {string} propertyName 
-   * @param {boolean | {(element: Element, name: string, newVal, oldVal) => any}} reflection 
+   * @param {{(element: Element, name: string, newVal, oldVal) => any}} instruction A function with suitable parameters to react for setting attribute on the element
    */
-  registerReflection(propertyName, reflection) {
+  _registerReflection(propertyName, instruction) {
     let reflections = this.reflections || (this.reflections = {});
     if (propertyName in reflections) {
       throw new Error(`Reflection for ${propertyName} was already registered`);
     }
-    if (typeof reflection === 'function') {
-      reflections[propertyName] = reflection;
-    } else if (reflection) {
-      reflections[propertyName] = propToAttr;
+    if (typeof instruction !== 'function') {
+      throw new Error('Invalid reflection instruction');
     }
+    reflections[propertyName] = instruction;
   }
   
   /**
    * @param {Element} element 
-   * @param {Function} target 
+   * @param {object} viewModel 
    */
-  _setupReflections(element, target) {
+  _setupReflections(element, viewModel) {
     if (!this.reflections) return;
-    if (this._reflectedTarget === target) return;
-
-    let {reflections} = this;
-    let method = 'propertyChanged';
-    let onChanged = target.prototype[method];
-    let hasHandler = !!onChanged;
-    
-    let alteredHandler = hasHandler
-      ? function propertyChanged(name, newVal, oldVal) {
-          onChanged.call(this, name, newVal, oldVal);
-          reflections[name].call(this, element, name, newVal, oldVal)
-        }
-      : function propertyChanged(name, newVal, oldVal) {
-          reflections[name].call(this, element, name, newVal, oldVal);
-        }
-
-    if (!Reflect.defineProperty(target.prototype, method, {
-      configurable: true,
-      value: alteredHandler
-    })) {
-      throw new Error(`Cannot setup property reflection on <${this.elementName}/> for ${target.name}`);
-    };
-    this._reflectedTarget = target;
+    let lookup = this.observerLocator.getOrCreateObserversLookup(viewModel);
+    lookup.__reflections__ = this.reflections;
+    lookup.__element__ = element;
   }
 
   _ensurePropertiesDefined(instance: Object, lookup: Object) {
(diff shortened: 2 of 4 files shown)


bigopon:
@AshleyGrant If the original request was to have the ability to reflect the prop to every element bound, not just the custom element itself, maybe it better suits a new binding command `attr` or a separate `bindingBehavior`:

```js
<div prop.attr='someThing'></div>

<div prop.bind='someThing & reflectToAttribute'></div>
```

src/bindable-property.js
```diff
@@ -35,7 +35,13 @@ export class BindableProperty {
   * Creates an instance of BindableProperty.
   * @param nameOrConfig The name of the property or a cofiguration object.
   */
-  constructor(nameOrConfig: string | Object) {
+  constructor(nameOrConfig: string | {
```
jods4: Thanks for typing that `Object`!
Should we have a named interface for that rather? `BindableOptions`? That would make the code cleaner + give a place to comment on members... I know that `attribute` does but it's not totally obvious, especially now with `reflect` in the mix.

src/bindable-property.js
```diff
@@ -35,7 +35,13 @@ export class BindableProperty {
...
+    defaultBindingMode?: number,
```
jods4: `number` :(
Don't we have an enum to match? Can we create one?

src/bindable-property.js
```diff
@@ -35,7 +35,13 @@ export class BindableProperty {
...
+    reflect?: boolean | {(el: Element, name: string, newVal, oldVal): any},
```
jods4: This is bike-shedding but I'm not a fan of the name. 
`reflect` is a little too generic for my taste, makes me think of `Reflection` first.
Polymer uses, `reflectToAttribute`, maybe we should play along? If someone sees this pop up in IntelliSense, its meaning is somewhat clear. `reflect` not as much.

jods4: I'm not convinced by the API design of the function part.
It feels like it takes too much... the element and old value and a name (of what)?

This is meant to just reflect a value in an attribute. The callback here really feels like `changeHandler`. Accepting a `bool | string` with the string allowing to reflect to a different attribute name seems simpler.

src/bindable-property.js
```diff
@@ -58,17 +64,55 @@ export class BindableProperty {
...
     if (descriptor) {
       this.descriptor = descriptor;
       return this._configureDescriptor(descriptor);
     }
 
     return undefined;
   }
+  
+  _configureReflection(target) {
+    if (target.__reflectionConfigured__) return;
```
jods4: Aurelia includes a `Symbol` polyfill.
Is the `__xxx__` field how we want to roll when hacking into objects we don't own? Shouldn't we use symbols instead?

src/bindable-property.js
```diff
@@ -58,17 +64,55 @@ export class BindableProperty {
...
+
+    let method = 'propertyChanged';
+    let onChanged = target.prototype[method];
+    let hasHandler = !!onChanged;
+    
+    let alteredHandler;
+    if (hasHandler) { // avoid ternary to make it consisten with the rest ?
+      alteredHandler = function propertyChanged(name, newVal, oldVal) {
+        onChanged.call(this, name, newVal, oldVal);
+        let { __element__, __reflections__ } = this.__observers__;
```
jods4: Not sure, so just asking: is accessing `__observers__` directly the way to go?
Isn't that an implementation detail of observation and an API should be used here rather?

AshleyGrant:
Great comments from @jods4!

src/bindable-property.js
```diff
@@ -58,17 +64,55 @@ export class BindableProperty {
...
+        if (!__element__) return;
+        __reflections__[name].call(this, __element__, name, newVal, oldVal);
+      }
+    }
+  
+    if (!Reflect.defineProperty(target.prototype, method, {
+      configurable: true,
+      value: alteredHandler
+    })) {
+      throw new Error(`Cannot setup property reflection on <${this.name}/> for ${target.name}`);
```
jods4: If we go for `throw` on failure, why don't use `Object.defineProperty`. Unlike `Reflect` it exists on older browsers, does the same thing, and throws when it fails.

src/bindable-property.js
```diff
@@ -248,3 +292,18 @@ export class BindableProperty {
...
+
+/**
+ * @private
+ * Used for avoid creating mapping function multiple times
+ * @param {Element} element 
+ * @param {string} propertyName 
+ * @param {any} newValue 
+ */
+function propToAttr(element, propertyName, newValue) {
+  if (newValue == null) {
```
jods4: I like this, but I think the rest of the team is linting for strict `===` always.

src/bindable-property.js
```diff
@@ -248,3 +292,18 @@ export class BindableProperty {
...
 }
+
+/**
+ * @private
+ * Used for avoid creating mapping function multiple times
+ * @param {Element} element 
+ * @param {string} propertyName 
+ * @param {any} newValue 
+ */
+function propToAttr(element, propertyName, newValue) {
```
jods4: It seems to me this doesn't take into account the camelCase to kebab-case between properties and html attributes? 

Also in HTML a few weird properties do not map directly to same-name attribute. I think we have a kind of map somewhere in Aurelia that converts some of those so that it "just work" when binding HTML attributes, shouldn't we re-use it here so that it just works for those?

Should this API interact with the `attribute` binding option? It allows users to map a property to a different name attribute, wouldn't they expect `reflectToAttribute` to take that into account? Seems more intuitive.

src/html-behavior.js
```diff
@@ -401,6 +405,33 @@ export class HtmlBehaviorResource {
...
+   * Allow a bindable property on custom element to register how to reflect prop value to attribute
+   * @param {string} propertyName 
+   * @param {{(element: Element, name: string, newVal, oldVal) => any}} instruction A function with suitable parameters to react for setting attribute on the element
+   */
+  _registerReflection(propertyName, instruction) {
+    let reflections = this.reflections || (this.reflections = {});
+    if (propertyName in reflections) {
+      throw new Error(`Reflection for ${propertyName} was already registered`);
+    }
+    if (typeof instruction !== 'function') {
```
jods4: Shouldn't parameter validation go first, before mutating `this`?

Judging by the `_` prefix, this is a private API, do we validate parameters on private APIs?

src/html-behavior.js
```diff
@@ -401,6 +405,33 @@ export class HtmlBehaviorResource {
...
+      throw new Error('Invalid reflection instruction');
+    }
+    reflections[propertyName] = instruction;
+  }
+  
+  /**
+   * @param {Element} element
```
jods4: Why use Typescript-like type annotations in `bindable-property.js` and JSDoc comments here?

jods4:
I had a quick look at the code. Not enough to validate the PR but the main mechanic looks good to me.

I am a little confused when you say it addresses aurelia/binding#347. What/how does it do that, I think I missed something?
Notably, one thing I really would like to be able to do is:
```html
<input magical-bool-binding />
```
As described in this comment https://github.com/aurelia/binding/issues/347#issuecomment-242106920

src/bindable-property.js
bigopon: This one is really silly of mine, was too hasty to see this. 👍 
How would you fix this though, there is already `getObserver` in the scope, in which it access `__observers__` directly. Maybe i should create `getObserverLookup` and point `getObserver` there. Or just leave `getObserver` alone to avoid performance cost of fn calling ?

src/bindable-property.js
bigopon: If we restricted it to only choose the name, whjat would we do if it is not always 1:1 prop->attr ?

src/bindable-property.js
bigopon: Some thing I couldn't think of. Can only 👍

src/bindable-property.js
jods4: Do you mean like one binding updates 2 attributes? For example `condition` boolean property is reflected into the presence of either an `true` or a `false` attribute on target element?

Personally I would say this is narrow use cases that can be filled today by using `changeHandler`, which has practically the same signature.

src/bindable-property.js
jods4: I did not see that `getObserver` function. I think the right thing to do is to use it to not break the abstraction level here.

If you want to maximize perfs and make sure `getObserver` gets optimized by JIT you can split it in two so that it is small and only contains the happy path (`lookup !== undefined`) and put the initialization code in a second function.

src/bindable-property.js
bigopon: No camelCase to kebab-case is a bad mistake. It should be fixed.

Code for the special pairs belong to `templating-binding`, should I copy it over ? I'm not sure how to handle this, because props of custom element in Aurelia, donot necessarily map to attribute like built-in props.

jods4: You are right. This is for custom attributes, it has nothing to do with built-in HTML attributes.
Forget my remark about the map ☺️

bigopon:
Thanks @jods4 , thorough as expected.

I understood of the original requirement was to reflect Aurelia custom element prop to attribute, so it needs some signal to know when the attribute should go. I was trying to do this by associating null / undefined value with the removal. But then I realise that if we are to support JS typing, and reflect a boolean property to prop, then we can provide two default reflecting behavior: 1:1 and boolean. Then by the setup time, bindable property will pickup the metadata and decide which to go. What you think ?

Overall i'm still not sure how you support flexibility here if we don't allow `reflectToAttribute` to be a function. If you could point out a solution, it would be great.

src/html-behavior.js
bigopon: It should be removed.

src/bindable-property.js
bigopon: For the `attribute`, it sounds really good. Maybe i'll go for it and update the PR doc.

src/bindable-property.js
bigopon: This will be fixed. Was trying to roll it out so a bit lazy for `||` and `===` 😄

src/bindable-property.js
bigopon: This is because my lack of knowledge about compiler result. I didn't know if `'use strict'` will always be there so i tried to do something.

jods4:
The thing about aurelia/binding#347 is that it's not just about updating attributes when the property changes.

It's also about having a shortcut notation where `<div x>` is shortcut for `<div x.one-time="true">`.

I think I don't fully grasp what use cases you want to address by allowing `reflectToAttribute` to be a function. Could you please provide a concrete example for the sake of discussion?

src/bindable-property.js
bigopon: I think I will move the whole block of reflection out of that function, to a separate fn

bigopon:
I totally missed that point of original request.

I don't have a use case for `reflectToAttribute` as a function. The motivation behind was just to have the flexibility when setting the value, as it's not always 1:1 prop->attr. Maybe `bindable` config could use another property for value conversion ?

src/bindable-property.js
bigopon: It's the case where value of property is an object, we dont want `[object Object]` in the attribute.

Another motivation, is that pretty much everything in Aurelia is intercept-able, So I would choose go with the rest on this.

jods4:
@bigopon let's see what others think of the proposed `reflectToAttribute`.

My opinion is that we shouldn't introduce complexity unless it has strong pay-off.

In this case:
- We don't have a good, concrete use case for it.
- It provides practically the same service as existing `changeHandler`.

So my vote would be to wait. Ship `reflectToAttribute: boolean` in v1 and add other options based on community feedback.

src/bindable-property.js
```diff
@@ -294,16 +298,25 @@ export class BindableProperty {
 }
 
 /**
+ * @param instance the view model instance
+ * @param propertyName name of property changed
+ * @param newValue
+ * @param oldValue
+ */
+function callRefelection(instance: Object, propertyName: string, newValue, oldValue) {
```
jods4: Typo here: _Refelection_

src/bindable-property.js
```diff
@@ -294,16 +298,25 @@ export class BindableProperty {
...
  * Used for avoid creating mapping function multiple times
- * @param {Element} element 
- * @param {string} propertyName 
- * @param {any} newValue 
  */
-function propToAttr(element, propertyName, newValue) {
-  if (newValue == null) {
-    element.removeAttribute(propertyName);
+function propToAttr(element: Element, propertyName: string, newValue: any) {
+  if (newValue === null || newValue === void 0) {
```
jods4: OK I might be nitpicking but I prefer `newValue === undefined`, for clarity.
Minifiers will do the job of using the smallest code.

I know users _could_ redefine `undefined` but a lot of other parts won't work if they do.

src/bindable-property.js
```diff
@@ -70,6 +87,37 @@ export class BindableProperty {
...
 
+  _configureReflection(target) {
+    if (target[reflectionConfigured]) return;
+
+    let method = 'propertyChanged';
+    let onChanged = target.prototype[method];
+    let hasHandler = !!onChanged;
+
+    let alteredHandler;
+    if (hasHandler) { // avoid ternary to make it consisten with the rest ?
```
jods4: I thought about this part some more...
It does something very similar to `createObserver` (same file).

Wouldn't it be more efficient and more clear to integrate `callReflection` inside `createObserver`?

jods4:
So this doesn't play nice with lifted custom elements (`@containerless`), because they don't have a target element to reflect to. 
Could we detect this situation and log an error in the console? 
Should only be on actual usage, I'm thinking of weird composition / inheritance cases.

bigopon:
@jods4 Thanks for your review, and I'm thinking of closing this PR. since it doesn't really add much value, like you suggested. (Wonder if you have thought of any scenarios this could be useful beside saving few lines of code).

And after realized I failed to address Ashley's case making me agree with you more. I feel like there is no point adding this once, as it is a bit complex to deal with inheritance and if anyone ever wanted this, it would just be a simple injection and reflect in `valueChanged` handler.

But I'd like to add that about your point of complex function as a value for `reflectToAttribute`, you are correct, maybe it should have the same signature with `valueConverters`, take `newValue`, `oldValue` and return something, then that will be mapped to the element, privately by the framework.

Maybe @EisenbergEffect can put a some thoughts before we close this PR ?

src/bindable-property.js
bigopon: I was trying to do minifier job, will avoid next PRs 👍

src/bindable-property.js
bigopon: 😢

jods4:
@bigopon I don't know why I gave you the impression that this PR "doesn't really add much value". I honestly never intended that.

I personally think that the _function parameter_ to `reflectToAttribute` adds too much burden for its value.

But as for `reflectToAttribute: boolean` it closes #283 so there's value in that.

I also like the extra typings that you put in this PR. It makes it much easier to work with Aurelia code. ❤️

jdanyow:
I think an `attr` command would be more generally applicable and easier for folks to use.
https://github.com/aurelia/templating-binding/issues/106#issuecomment-302850701

Before this is closed, are there any use-cases or pain points that this `@bindable({ reflect: true })` approach solves that can't be solved with `foo.attr="something"` ?

EisenbergEffect:
Part of this addresses the issue with making bindable properties have the same capability as built-in elements, the potential to automatically reflect property changes to attributes. This includes boolean attributes (bindables that add/remove the attribute from the element altogether), which we need for the UX library.

If we add the attr command, that would need to work with bindable properties, since there's little reason to create a bindable where you don't want the property to be updated. So, it's not just an attr update in this case.

Beyond that, attr reflection should be independent of one-way or two-way capability. I'm not sure how this would be handled, even though it's a rare case.

jods4:
Also: this feature is copied from Polymer.
Polymer is meant for building/using web components and as Rob said, it's kind of standard for elements (built-in or custom) to reflect (some of) their properties into attributes.

jdanyow:
Apologies- when I was reading the discussion I got the impression that @bigopon was closing this.

Still not clear on the use-case though. I see polymer has it, what's the purpose in the aurelia context? To enable targeting states with css attribute selectors? What's the aurelia-ux use-case?

EisenbergEffect:
Yes, that's definitely one of them.

bigopon:
I was thinking of it, as I'm not sure if it's worth the lines. @jdanyow 

Also I was concerned by the fact that there may be confusion in the `bindable`, if we support `attribute` property in `bindable` config, as some will try to use `attribute` as some sort of how to reflect it back to the element. While some others will not touch it and think it as an instruction for custom attribute only.

Maybe we can have 

```ts

type ReflectInstruction = string | {
  attr: string,
  converter: (newValue, oldValue) => any
}

reflectToAttribute: ReflectInstruction
```

bigopon:
I've followed @jods4 's suggestion.

```js
  @bindable({
    reflectToAttribute: 'boolean' | 'string', // also support (element, attrName, value) as the reason given with object -> [object Object]. Or should it just be gone ?
    attribute: 'given-name' // this means prop firstName will be reflected to given-name
  })
  firstName
```

jdanyow:
Does this work in both directions? If someone calls `setAttribute`, will the associated bindable property update?

bigopon:
Is it this feature request templating#458 you are referencing ? It's would be nice to have, but @EisenbergEffect didn't give any comment about resolution. I think you commented somewhere in some related issue about using `MutationObserver` for that.

danfma:
Hey guys, is this feature being verified?

PR #564, opened 2017-09-24
feat(Let): let element for computed property declaration

bigopon:
[Base discussion](https://github.com/aurelia/templating-binding/issues/115)

This is a step 1 to support let element.
Step 2 will be in `templating-binding` with implementation of let expression classes, and `createLetExpressions` for binding language.

requires aurelia/templating-binding#117

#### Usage

```html
<template>
  <let
    msg.bind='message'
    fun-message.bind='msg | fun'
    no-fun-message.bind='msg | noFun'
    sentence='Full name is: "${fullName}" and first name is "${firstName}" and last name is "${lastName}"'
    literal-string='I am a literal stirng in let element. Triggered warning.'></let>

  <!-- compute dayName on each loop bindingContext -->
  <div repeat.for='day of days'>
    <let day-name.bind='day | toShort'></let>
    <span>Day name: ${dayName}. Index: ${$index}</span>
  </div>
</template>
```

### DIFF
--- a/src/view-factory.js
+++ b/src/view-factory.js
@@ -428,6 +428,7 @@ export class ViewFactory {
     let fragment = createInstruction.enhance ? this.template : this.template.cloneNode(true);
     let instructables = fragment.querySelectorAll('.au-target');
     let instructions = this.instructions;
+    let letExpressions = instructions.letExpressions;
     let resources = this.resources;
     let controllers = [];
     let bindings = [];
@@ -442,6 +443,10 @@ export class ViewFactory {
     let instruction;
 
     this.resources._invokeHook('beforeCreate', this, container, fragment, createInstruction);
+
+    for (i = 0, ii = letExpressions.length; ii > i; ++i) {
+      bindings.push(letExpressions[i].createBinding());
+    }
 
     if (element && this.surrogateInstruction !== null) {
       applySurrogateInstruction(container, element, this.surrogateInstruction, controllers, bindings, children);
--- a/src/view-factory.js
+++ b/src/view-factory.js
@@ -169,7 +169,7 @@ function applyInstructions(containers, element, instruction, controllers, bindin
     return;
   }
 
-  if (instruction.letExpressions) {
+  if (instruction.letElement) {
     for (i = 0, ii = expressions.length; i < ii; ++i) {
       bindings.push(expressions[i].createBinding());
     }
@@ -447,7 +447,6 @@ export class ViewFactory {
     let fragment = createInstruction.enhance ? this.template : this.template.cloneNode(true);
     let instructables = fragment.querySelectorAll('.au-target');
     let instructions = this.instructions;
-    let letExpressions = instructions.letExpressions;
     let resources = this.resources;
     let controllers = [];
     let bindings = [];
@@ -462,10 +461,6 @@ export class ViewFactory {
     let instruction;
 
     this.resources._invokeHook('beforeCreate', this, container, fragment, createInstruction);
-
-    for (i = 0, ii = letExpressions.length; ii > i; ++i) {
-      bindings.push(letExpressions[i].createBinding());
-    }
 
     if (element && this.surrogateInstruction !== null) {
       applySurrogateInstruction(container, element, this.surrogateInstruction, controllers, bindings, children);
--- a/src/view-factory.js
+++ b/src/view-factory.js
@@ -126,6 +126,17 @@ function makeElementIntoAnchor(element, elementInstruction) {
   return anchor;
 }
 
+/**
+ * @param {Container[]} containers
+ * @param {Element} element
+ * @param {TargetInstruction} instruction
+ * @param {Controller[]} controllers
+ * @param {Binding[]} bindings
+ * @param {ViewNode[]} children
+ * @param {Record<string, ShadowSlot>} shadowSlots
+ * @param {Record<string, ViewFactory>} partReplacements
+ * @param {ViewResources} resources
+ */
 function applyInstructions(containers, element, instruction, controllers, bindings, children, shadowSlots, partReplacements, resources) {
   let behaviorInstructions = instruction.behaviorInstructions;
   let expressions = instruction.expressions;
@@ -155,6 +166,14 @@ function applyInstructions(containers, element, instruction, controllers, bindin
     DOM.replaceNode(commentAnchor, element);
     shadowSlots[instruction.slotName] = slot;
     controllers.push(slot);
+    return;
+  }
+
+  if (instruction.letExpressions) {
+    for (i = 0, ii = expressions.length; i < ii; ++i) {
+      bindings.push(expressions[i].createBinding());
+    }
+    element.parentNode.removeChild(element);
     return;
   }
 
--- a/dist/aurelia-templating.d.ts
+++ b/dist/aurelia-templating.d.ts
@@ -22,6 +22,7 @@ import {
   inject
 } from 'aurelia-dependency-injection';
 import {
+  Scope,
   ValueConverterResource,
   BindingBehaviorResource,
   camelCase,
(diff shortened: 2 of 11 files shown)


bigopon:
I think the fix at aurelia/binding#622 should be considered to be released with this new feature. As there will be more subscribers floating around with the introduction of let.

cc @EisenbergEffect @jdanyow

jdanyow:
@bigopon  re: https://github.com/aurelia/templating/pull/564#issuecomment-331734083 :  sounds good, I've posted a question on that PR

bigopon:
@jdanyow Was trying to make the job easier for you, but couldn't find the appropriate place to write doc 😅 . Added typing for new changes though

EisenbergEffect:
Looks like this has stalled out. Besides rebasing on master to address the conflicts, is this something that is ready?

bigopon:
Let me put down some doc and it should be good, beside rebasing. Let me review it again, i think code around bindingcontext and override context can be simplified. I was hoping for some solid review from @jdanyow and @EisenbergEffect too

bigopon:
@jdanyow @EisenbergEffect 

Doc added. It's ready now. The gist is updated with latest changes in `templating` for demo at https://gist.run/?id=80d070452f6c5afbe9173571422da317

RomkeVdMeulen:
Stalled again?

Alexander-Taran:
Having separation of concerns issues with this.
looks like declaring a variable in view for the vm
my 2c

glen-84:
@Alexander-Taran,

Twig has a [set tag](https://twig.symfony.com/doc/1.x/tags/set.html), and it's quite useful.

If you're applying some VCs to a value and then using the result in multiple places, or even just using it to reduce line length:

```html
<deeply-nested>
    ${property | transform | something:param1 | pluck:'blah' & signal:'some-signal'}
</deeply-nested>
```

Here, the `deeply-nested` element is already at, for example, column 110, when your recommended line length is 120 characters, so this long usage of VCs and signallers exceeds that limit.

```html
<!-- This can happen at a less indented level -->
<let propTransformed.bind="property | transform | something:param1 | pluck:'blah' & signal:'some-signal'"></let>
<deeply-nested>
    ${propTransformed}
</deeply-nested>
<another-usage>
    ${propTransformed}
</another-usage>
```

Alexander-Taran:
@glen-84 I get it. I really do. I just don't like the feel of it leaking back to vm.
It "should" belong in vm IMO

fkleuver:
I think this is a good idea. Would not encourage people to do this everywhere, but occasionally this can let you reduce many lines of viewmodel code into a simple one-liner in your view. There's no easy way to use converters and behaviors in viewmodels after all..

bigopon:
Update: I think syntax can be more flexible, so @Alexander-Taran doesn't get grumpy. By default it will assign value to `overrideContext` to have it available in view **only**, only assign to view model when specified explicitly

```html
<!-- stays in the view -->
<let
  filtered-items.bind="items | filter :criteria1 :criteria2 | sort :criteria11 :criteria12 "
  ></let>

<!-- goes to view model -->
<let
  to-view-model
  filtered-items.bind="items | filter :criteria1 :criteria2 | sort :criteria11 :criteria12 "
  ></let>
```

Thought ? @EisenbergEffect

EisenbergEffect:
@bigopon I'd be interested in seeing an implementation of this as described. However, I have an extra requirement :) You have to submit an equivalent implementation to the experiment repo for vNext.

Alexander-Taran:
A suggestion to consider: `filtered-items.view-only /.to-view-model | .crazilyMakeItAvailableInVMasWell`

glen-84:
Is `to-view-model` even needed? Are there use cases?

For my (current) use cases, view-only is sufficient.

bigopon:
Assigning to view model gives a way to react to change, which is useful when binding expressions contain value converters

bigopon:
https://github.com/aurelia/templating/blob/c942f083de027bc4ed0d7aa07a3f4b3276b1a448/src/view-compiler.js#L114

@EisenbergEffect I'm using this to store `<let/>` instruction, it will make a difference between vCurrent and vNext, as vCurrent disregard location of `<let/>` while vNext does not. Any comment here ?

EisenbergEffect:
I'd rather not introduce `let` into vCurrent if it's going to immediately cause a breaking change with how we've already implemented it in vNext.

src/view-factory.js
```diff
@@ -158,6 +169,14 @@ function applyInstructions(containers, element, instruction, controllers, bindin
     return;
   }
 
+  if (instruction.letElement) {
```
bigopon: @EisenbergEffect Here is the difference compared to the previous implementation, it will just be a normal `TargetInstruction` with `.letElement === true`  and thus, behaves the way it's specified in the DOM.

EisenbergEffect:
Ok, do you feel this is ready to release? Also, do you feel confident this is fully forward compatible with vNext? Just want to be extra careful ;)

bigopon:
@EisenbergEffect, i will need to adjust few details / tests in `templating-binding`, the `templating` side is ready.

Re: compat, as far as i could see when doing both, they are compatible, string literal in vnext is a hard error instead of a warning like vcurrent but its temporary. Should we have a spec document to define the behavior for both version, and then expand to the rest of the framework? I think its implementation based so it makes less sense but it could help.

bigopon:
@EisenbergEffect Theoretically only `templating-binding` depends on `templating`. So `templating` new version can be upped without `templating-binding`. There are 2 scenarios:
  * Apps without `<let/>` element in their global resources will continue work fine, with only new version of `templating`.
  * App with `<let/>` in their global resources (the one copied from @jdanyow example gist), will break without upgrading `templating-binding` together with `templating`. (implementation wise, the gist and my PR are the same, with extra `to-view-model` in my PR, and more efficient). For this, my take is to have a compromise here and release it as is. If folks complain, we can do a follow up PR to do a pre-check to see if there is `<let/>` in the resources during compilation and treat `<let/>` like normal custom element.

bigopon:
> App with <let/> in their global resources (the one copied from @jdanyow example gist), will break without upgrading templating-binding together with templating. (implementation wise, the gist and my PR are the same, with extra to-view-model in my PR, and more efficient). For this, my take is to have a compromise here and release it as is. If folks complain, we can do a follow up PR to do a pre-check to see if there is <let/> in the resources during compilation and treat <let/> like normal custom element.

Maybe we can do it from the start to guarantee this is an absolutely safe addon. @EisenbergEffect

EisenbergEffect:
@bigopon Can you expand on that last idea? How much work do you think that would be and would it effect startup time, for example?

bigopon:
@EisenbergEffect At the moment for this PR, this is the check:
```js
  if (tagName === 'slot') {
    // ...
  } else if (tagName === 'let') {
    // compile let expression
  } else if (tagName === 'template') {
    // ...
  } else {
    // ...
  }
```
It's unsafe because it potentially break folks app who has `<let/>` custom element.
We can change it to following:
```js
  if (tagName === 'slot') {
    // ...
  } else if (tagName === 'template') {
    // ...
  } else {
    customElementType = resources.getElement(tagName);
    if (tagName === 'let' && customElementType === null) {
      // this means it's safe to employ our default implementation for <let/>
      // can go even safer, to make sure no app breaks:
      if (bindingLanguage.compileLetExpressions !== defaultNoop) {
        // defaultNoop will not be like other binding language abstract methods, wont throw
        // this branch means app has a binding language with proper compileLetExpression, safe to process
        // process <let/>
        return;
      }
    }
    // all the check above done, means it's just a normal custom element, potentially `<let/>` custom element, or a normal <let/>, without requiring binding language to be updated at the same time with templating
  }
```

EisenbergEffect:
💯 Let's do this @bigopon I like this approach.

fkleuver:
I like it too. Very thorough in making sure no existing apps break :)

bigopon:
@EisenbergEffect The implementations are done on both `templating` and `templating-binding`

bigopon:
For release note, there should be a note to remind devs to update both `templating` and `templating-binding` to have `<let/>` working.

PR #567, opened 2017-10-08
fix(view-engine): define metadata for SlotCustomAttribute

### LINKED ISSUE #566: Error when having `slot` attribute on `<slot>`

* **Library Version:**
aurelia-templating 1.5.0

* **Operating System:**
OSX 10.12.6

* **Node Version:**
7.10.1

* **NPM Version:**
4.2.0

* **Browser:**
Chrome 63.0.3232.0

* **Language:**
TypeScript 2.5

**Current behavior:**
Error `TypeError: Cannot read property 'isInitialized' of undefined` when projecting a slot into another slot, i.e.

```html
<mixed-slot>
    <slot name="slot1" slot="slot1">Fallback Content for Projected Slot 1</slot>
    <slot name="slot2" slot="fallbackSlot2">Fallback Content for Projected Slot 2</slot>
</mixed-slot>
```

Reproduced in [niclassahlin/aurelia-templating-bug](https://github.com/niclassahlin/aurelia-templating-bug).

StrahilKazlachev:
fix #566
@EisenbergEffect I have some concerns about the fix, since `ViewEngine` *creates* and *initializes* the behavior instance:
- currently the relation is `au-slot` behavior instance per `ViewEngine` instance.
- defining the metadata changes this - now it's on the ` class SlotCustomAttribute`.
    - is it supported use case to have more than one instance of `ViewEngine` in an app(`enhance`)?
    - is it supported use case to have more than one apps running in parallel?

### DIFF
--- a/src/view-engine.js
+++ b/src/view-engine.js
@@ -75,6 +75,7 @@ export class ViewEngine {
 
     let auSlotBehavior = new HtmlBehaviorResource();
     auSlotBehavior.attributeName = 'au-slot';
+    metadata.define(metadata.resource, auSlotBehavior, SlotCustomAttribute);
     auSlotBehavior.initialize(container, SlotCustomAttribute);
     auSlotBehavior.register(appResources);
   }


EisenbergEffect:
I think this will work. I'm going to move the first three lines out of the instance so the same behavior would be shared. Init is guarded so that it's only run once.

PR #582, opened 2017-11-29
perf(child-observation) handle fast special case for wildcard (*) selector

rmja:
In the case with a significant portion of children (>2000), the `n+1` computation of the splice index becomes a significant overhead. This fix bypasses this computation in the special case when the selector is *.

I was unable to reliably build the dist output - it seems to have too many changes...

Before:
before

After:
after

### DIFF
--- a/src/child-observation.js
+++ b/src/child-observation.js
@@ -238,6 +238,12 @@ class ChildObserverBinder {
 
       if (this.all) {
         let items = (this.viewModel[this.property] || (this.viewModel[this.property] = []));
+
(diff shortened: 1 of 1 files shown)


rmja:
Is there anything I can do on my side to get this merged - it reduces load times on some of our pages with roughly half a second, cc @EisenbergEffect

EisenbergEffect:
Sorry for the delay. We'll look to get this into a release this weekend or early next week. Thanks for the contribution!

PR #584, opened 2017-12-09
refacor(ElementEvents): avoid new fn

bigopon:


bigopon:
@EisenbergEffect Maybe consider aliasing the export as well, as `ElementEvents` is not a correct name for this class. It can work with any subclass of `EventTarget`.

Also it can be extended to have a counter, like `subscribeCount(event, handler, capture, counter = -1)`

src/element-events.js
```diff
@@ -48,42 +49,25 @@ export class ElementEvents {
...
-      handler.handler = handler;
-      handler.bubbles = bubbles;
-      handler.dispose = () => {
-        this.element.removeEventListener(eventName, handler, bubbles);
-        this._dequeueHandler(handler);
-      };
-      this.element.addEventListener(eventName, handler, bubbles);
-      this._enqueueHandler(handler);
-      return handler;
+  subscribe(eventName: string, handler: Function, captureOrOptions?: boolean | AddEventListenerOptions = true): EventHandler {
```
EisenbergEffect: This looks like a breaking change to the third parameter. Now the default would be the opposite of what it was before, correct?

bigopon: From the naming of the parameter, yes. From the code, no. It would be considered a bug from the naming imo

EisenbergEffect: That sounds good then.

PR #586, opened 2017-12-14
feat(ElementEvents): batch subscription

bigopon:
This PR enhance `ElementEvents` class with ability to subscribe to multiple events at once.

Usage example:

```js
  import { ElementEvents } from 'aurelia-framework'

  const input = document.createElement('input');
  const elementEvents = new ElementEvents(input);
  elementEvents.subscribe({
    input: () => {
        // handle input value changed
    },
    blur: () => {
        // handle input blur
    },
    focus: {
      handler: () => {
        // handle input focus
      },
      once: true
    }
  });
```

Also compatible with `subscribeOnce`

```js
  import { ElementEvents } from 'aurelia-framework'

  const input = document.createElement('input');
  const elementEvents = new ElementEvents(input);
  elementEvents.subscribeOnce({
    input: () => {
        // handle input value changed
    },
    blur: () => {
        // handle input blur
    },
    focus: {
      handler: () => {
        // handle input focus
      },
      once: false
    },
    keydown: {
      handler: () => {},
      once: true,
      capture: true
    }
  });
```

The result of `subscribe` / `subscribeOnce` will be an plain object with same properties with input:

```js
  import { ElementEvents } from 'aurelia-framework'

  const input = document.createElement('input');
  const elementEvents = new ElementEvents(input);
  const subscriptions = elementEvents.subscribe({
    input: () => { },
    blur: () => { },
    focus: {
      handler: () => { },
      once: false
    }
  });

  Object.keys(subscriptions); // <======= ['input', 'blur', 'focus']
```

### DIFF
--- a/dist/aurelia-templating.d.ts
+++ b/dist/aurelia-templating.d.ts
@@ -34,12 +34,6 @@ import {
 import {
   TaskQueue
 } from 'aurelia-task-queue';
-export declare interface EventHandler {
-  eventName: string;
-  bubbles: boolean;
-  dispose: Function;
-  handler: Function;
-}
 
 /**
 * Specifies how a view should be created.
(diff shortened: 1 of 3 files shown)


bigopon:
If this is merged, then it should be encouraged over the other overload signature with tradition arguments:

```js
subscribe(
  eventName: string,
  handler: Function,
  captureOrOptions?: boolean | AddEventListenerOption): EventHandler;
```

as the third parameter value is `true`, which is not usually expected.

dist/aurelia-templating.d.ts
```diff
@@ -472,10 +466,29 @@ export declare class ViewEngineHooksResource {
...
+  eventName: string;
+  bubbles: boolean;
+  dispose: Function;
+  handler: Function;
+}
+
+export declare interface SubscriptionHandlerConfig {
+  handler: Function
+  capture?: boolean
+  passive?: boolean
```
zewa666: Whats this passive prop for? I see it nowhere handled in code

bigopon: It is used when we pass the whole object to `EventTarget.prototype.addEventListener` third paramter. Same with `passive` in spec.

Vheissu:
Would love to see this one merged, this is fantastic.

stsje:
How often do you actually need this? Usually you would subscribe to events in the view `<button click.delegate="">`

I would keep the API as it is, and not introduce a change with minimal gains

PR #587, opened 2017-12-15
feat(HtmlBehavior): new 'assigned' lifecycle

### LINKED ISSUE #257: Call `changed` callbacks after all bound properties are assigned during the initial bind phase
This is a request to change the timing of when `changed` callbacks are called during the `bind()` phase.

**Current timing:** The `changed` callbacks are invoked progressively, as the bound properties are set. The order in which they are called is not guaranteed.

Therefore this code does not work because in one `changed` callback the _other_ property will not be bound yet:

```
@bindable attr1;
@bindable attr2;
attr1Changed(){
   this.DoSometingWith(this.attr1, this.attr2);
}
attr2Changed(){
   this.DoSometingWith(this.attr1, this.attr2);
}
```

**Proposed timing:** The `changed` callbacks should be invoked after all bound properties are set.

I have already discussed this with @EisenbergEffect.

bigopon:
Introduce new `assigned` life cycle, which when presented, will be called after all `@bindable` properties have been bound. This lifecycle happens before bind, and presence of bind does not affect it. example:

```js
export class TextBox {
  @bindable value
  @bindable label

  assigned() {
    
  }
}
```

resolve #257

@EisenbergEffect @jdanyow

### DIFF
--- a/src/controller.js
+++ b/src/controller.js
@@ -86,6 +86,7 @@ export class Controller {
   * @param scope The binding scope.
   */
   bind(scope: Object): void {
+    let handlesAssigned = this.behavior.handlesAssigned;
     let skipSelfSubscriber = this.behavior.handlesBind;
     let boundProperties = this.boundProperties;
     let i;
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
I like the idea of creating a new hook so that we don't break things. I'm not 100% convinced this is what we should name it though. Any other ideas?

bigopon:
It was quite challenging for me to come up with a good name, others were:

* `bound`
* `changed`
* `inited`
* `initValue`
* `initedBindable`

Alexander-Taran:
Now pretending that I'm a complete newcomer:

How do i pick one? 
What should I use? bind or assigned? And what are those valueChanged stufff.. Maaaaaan!!!

This are not real questions, but empathy exercice.
There really got to be a reason for such confusion.

// Here be Trolling. If one would try to answer my questions above.

EisenbergEffect:
I *think* we'll likely hold on this until vNext and then introduce a breaking change where  a `bound` callback is consistently called after all properties are bound. Based on our experimental work, this approach actually would be pretty simple to implement.

EisenbergEffect:
Closing this. vNext now calls `bound` after all properties are bound.

PR #589, opened 2017-12-18
ViewCompiler: remove binding commands after compilation

bigopon:
It's an official way to clean up aurelia view. This mostly benefits `repeat`, the gain is small / maybe negative. example:

before:

```html
<input
  class="au-target"
  show.bind="condition"
  value.bind="..."
  focus.bind="..."
  au-target-id="1"/>
```

after

```html
  <input class="au-target" au-target-id="1"/>
```

Plugin that gives this functionality: https://github.com/lgabeskiria/aurelia-clean-bindings

benchmark at https://gist.run/?id=d4c665917deed8913b79bdc675c28cc3

### DIFF
--- a/src/view-compiler.js
+++ b/src/view-compiler.js
@@ -118,6 +118,7 @@ export class ViewCompiler {
       factory.setCacheSize(cacheSize);
     }
 
+    this._removeCommands(factory.template);
     resources._invokeHook('afterCompile', factory);
 
     return factory;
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Let's leave this as a plugin rather than built in.

Alexander-Taran:
@bigopon would be nice to move that one to contrib (-:

bigopon:
@Alexander-Taran Yes, but I'm not the author of that plugin :D

Alexander-Taran:
@bigopon I know (-:

doktordirk:
can add it here https://github.com/aurelia-contrib/awesome-aurelia at least

Issue #541, opened 2017-04-06
unbind not invoked for custom-elements under root level repeat.for within slot content


AdamWillden:
**I'm submitting a bug report**

* **Library Version:**
1.4.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
7.50

* **NPM Version:**
4.2.0

* **JSPM OR Webpack AND Version**
JSPM 0.17.0@beta.41

* **Browser:**
Chrome

* **Language:**
TypeScript 2.2

**Current behavior:**
This is a very obscure bug, and is best explained with an example. I'm not sure if the `repeat.for` is part of the root issue here or not.

```html
  <collapsible-island>
    <h3 slot="header">${dmaFunction}</h3>

    <div repeat.for="loggerLink of loggerLinks">
      <dma-mini-status if.bind="isImport" dma-id.bind="loggerLink.linkedDmaId"></dma-mini-status>
      <dma-status-logger logger-link.bind="loggerLink"></dma-status-logger>
      <dma-mini-status if.bind="isExport" dma-id.bind="loggerLink.linkedDmaId"></dma-mini-status>
    </div>
  </collapsible-island>
```

Our custom elements `<dma-mini-status>` and `<dma-status-logger>` components have bind called on render but `unbind` is never called.

However if we wrap the default content in a `<div>` (as shown below) then everything works as normal.
```html
  <collapsible-island>
    <h3 slot="header">${dmaFunction}</h3>

    <div>
      <div repeat.for="loggerLink of loggerLinks">
        <dma-mini-status if.bind="isImport" dma-id.bind="loggerLink.linkedDmaId"></dma-mini-status>
        <dma-status-logger logger-link.bind="loggerLink"></dma-status-logger>
        <dma-mini-status if.bind="isExport" dma-id.bind="loggerLink.linkedDmaId"></dma-mini-status>
      </div>
    </div>
  </collapsible-island>
```

I'm a bit rushed now so a reproduction is difficult for me to produce right now but will try to follow up soon.

It appears @RWOverdijk and @t03apt are actually experiencing the same issue:  https://gitter.im/SpoonX/Dev?at=58e630e468bee3091f17068f

StrahilKazlachev:
Does [this](https://gist.run/?id=2709c78ff902ad1fea6bce910bf66dae) represent your case?

AdamWillden:
[...]

PR #592, opened 2017-12-24
fix(ShadowDOM): respect `returnToCache` flag

### LINKED ISSUE #590: Problem with if.bind + repeat.for + @children selector

* **Library Version:**
1.1.5

* **Operating System:**
Windows 10

* **Node Version:**
8.4.0

* **NPM Version:**
5.30

* **JSPM OR Webpack AND Version**
webpack 3.5.5

* **Browser:**
All

* **Language:**
TypeScript 2

**Current behavior:**

When we have a situation like this:

```
    <test if.bind="visible">
        <test-child repeat.for="point of points"></test-child>
    </test>
```
and the `test` custom element has a property with the `@children("test-child")` decorator.

When we toggle the status of the visible variable (so the repeat.for is detached and attached again) we get the children array update notification but the children array has the wrong number of elements (it grows without limits each toggle, it seems that the new elements are added twice)

See this gist with console open:

https://gist.run/?id=923ce4224412a695890e8d134955c20c

Click on the toggle button. Notice how the array count increases by 3 to infinity without reason, even if the actual custom elements remain 3.

The array length should be equal to the children element count.

bigopon:
fixes #590 , #541

Previously `returnToCache` flag was ignored when removing projected content, causing views to be left bound even removed.

@EisenbergEffect @jdanyow 

Edit:

Related:

- https://github.com/aurelia/templating-resources/issues/310

### DIFF
--- a/src/view-slot.js
+++ b/src/view-slot.js
@@ -455,6 +455,9 @@ export class ViewSlot {
     if (this.isAttached) {
       view.detached();
     }
+    if (returnToCache) {
+      view.returnToCache();
+    }
   }
 
   _projectionRemoveAt(index, returnToCache) {
@@ -466,6 +469,9 @@ export class ViewSlot {
     if (this.isAttached) {
       view.detached();
     }
+    if (returnToCache) {
+      view.returnToCache();
+    }
   }
 
   _projectionRemoveMany(viewsToRemove, returnToCache?) {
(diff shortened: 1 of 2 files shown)


bigopon:
fix gist https://gist.run/?id=bc846c471e1b2925fb86e7c51f9d154f

PR #598, opened 2018-01-23
fix(BehaviorPropertyObserver): use Object.is() for comparison

bigopon:
* requires aurelia/polyfills#58
* resolves aurelia/binding#613

Performance shouldn't take any hit with native `Object.is` implementation, app using polyfill of `Object.is` will. Overall won't be a big deal and Aurelia will be more reliable as comparison between `NaN`, `+0`, `-0`  will no longer an issue.

### DIFF
--- a/src/behavior-property-observer.js
+++ b/src/behavior-property-observer.js
@@ -38,7 +38,7 @@ export class BehaviorPropertyObserver {
   setValue(newValue: any): void {
     let oldValue = this.currentValue;
 
-    if (oldValue !== newValue) {
+    if (Object.is(newValue, oldValue)) {
       this.oldValue = oldValue;
       this.currentValue = newValue;
 
@@ -62,7 +62,7 @@ export class BehaviorPropertyObserver {
 
     this.notqueued = true;
 
-    if (newValue === oldValue) {
+    if (Object.is(newValue, oldValue)) {
       return;
     }
 


bigopon:
@EisenbergEffect

EisenbergEffect:
@bigopon Can you remind me which browsers will cause the polyfill to kick in for the above scenario?

EisenbergEffect:
Seems like it only affects IE, which is fine with me. Will get this in shortly!

bigopon:
According to this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is

It should be also Safari < 9

bigopon:
According to this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is

It should be also Safari < 9

@EisenbergEffect

src/behavior-property-observer.js
```diff
@@ -38,7 +38,7 @@ export class BehaviorPropertyObserver {
   setValue(newValue: any): void {
     let oldValue = this.currentValue;
 
-    if (oldValue !== newValue) {
+    if (Object.is(newValue, oldValue)) {
```
EisenbergEffect: @bigopon Should there be a `!` in front of this?

bigopon: Yes, it's fixed. Thanks for catching that.

PR #601, opened 2018-02-13
feat(interfaces): define and specify a well typed configuration object

aluanhaddad:


aluanhaddad:
I just realized this overlaps with https://github.com/aurelia/templating/pull/560, albeit in a non-conflicting way.

StrahilKazlachev:
@aluanhaddad  can you exclude the dist, or did you add it to illustrate something?
Also the other and this PR are missing `primaryProperty?: boolean`(attributes only) in the config.
@bigopon what do you think about adding the typings now? Or is #560 close to being complete?

bigopon:
It's @EisenbergEffect 's / @jdanyow 's call, the other PR will take sometime to land

src/interfaces.js
```diff
@@ -67,3 +68,21 @@ interface DynamicComponentGetViewStrategy {
     */
     getViewStrategy(): string|ViewStrategy;
 }
+
+/**
+* An optional interface describing the configuration object that can be specified to customize bindable properties.
+*/
+interface BindablePropertyConfig {
```
bigopon: it also takes a property name `defaultValue`, type any

aluanhaddad: Thanks!

aluanhaddad:
Thanks @bigopon and @StrahilKazlachev!

I have made the requested changes.

I didn't know about either attribute. `primaryProperty` looks awesome, super useful.

src/decorators.js
```diff
@@ -58,7 +58,7 @@ export function customElement(name: string): any {
 * @param defaultBindingMode The default binding mode to use when the attribute is bound with .bind.
 * @param aliases The array of aliases to associate to the custom attribute.
 */
-export function customAttribute(name: string, defaultBindingMode?: number, aliases?: string[]): any {
+export function customAttribute(name: string, defaultBindingMode?: bindingMode, aliases?: string[]): any {
```
EisenbergEffect: `bindingMode` is unfortunately a const object at this time. So, I don't think it can be used as a type. Can we revert this back to number?

src/interfaces.js
```diff
@@ -67,3 +68,29 @@ interface DynamicComponentGetViewStrategy {
...
+  /**
+  * The default binding mode of the property.
+  */
+  defaultBindingMode?: bindingMode;
```
EisenbergEffect: I think this will need to be number, since bindingMode is actually an object, not an enum or union type.

Alexander-Taran: It's not that simple.
From usage it should be union type.
But I can't see how to make it possible

EisenbergEffect:
It’s not actually TypeScript, so we can’t use a union or an enum. We have
to make it a number unfortunately.


On July 5, 2018 at 4:30:45 PM, Alexander Taran (notifications@github.com)
wrote:

*@Alexander-Taran* commented on this pull request.
------------------------------

In src/interfaces.js
<https://github.com/aurelia/templating/pull/601#discussion_r200515848>:

> @@ -67,3 +68,29 @@ interface DynamicComponentGetViewStrategy {
     */
     getViewStrategy(): string|ViewStrategy;
 }
+
+/**
+* An optional interface describing the configuration object that can
be specified to customize bindable properties.
+*/
+interface BindablePropertyConfig {
+  /**
+  * The default binding mode of the property.
+  */
+  defaultBindingMode?: bindingMode;

It's not that simple.
From usage it should be union type.
But I can't see how to make it possible

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/aurelia/templating/pull/601#discussion_r200515848>, or mute
the thread

src/decorators.js
bigopon: Note that with newest API on `BindableProperty`, it now understands lookup property of `bindingMode` such as `toView`, `fromView`, `twoView`, `oneTime`

@EisenbergEffect do we need to adjust this based on that ?

EisenbergEffect: @bigopon So, it will accept either a string or a number?

bigopon: @EisenbergEffect Yes it's added in this commit https://github.com/aurelia/templating/commit/e7da9730eaafb6c9522a659fa8355f0f49188e56#diff-8fecf3f03eea2e940742fe22c0791bfbR51 

It was to support Aurelia plugins / classes

PR #602, opened 2018-02-24
doc(templating-basics): Added if-else template controller to docs

pwoosam:
Added description with sample code of if-else binding using Aurelia.


EisenbergEffect:
Awesome! Thanks!

PR #603, opened 2018-02-24
doc(templating-basics): Added a note on using repeat.for with arrays

pwoosam:
Added a note on using repeat.for with arrays.

Use the array methods instead of replacing array elements by index with `[]` if you want Aurelia to observe the changes.


EisenbergEffect:
Another good addition. Thanks!

PR #606, opened 2018-03-25
feat(StaticViewStrategy): ability to statically analyze aurelia modules

bigopon:
This PR adds static view strategy, which will give ability to declare all resources used in a view statically, also optionally lazily.

## Syntax:

```js
@staticView({
  // the template of this custom element, can be omitted if custom attribute
  template: HTMLTemplateElement | null | undefined;

  // the resources that will be used inside the view
  resources: Function[];

  // address of the view, in case there is some dynamic loading involved
  moduleId?: string;

  // used to declare lazy dependencies, mainly for code splitting purposes
  imports?: () => Promise<Record<string, Function>>;
})
```

## Usage example:

```js
import { CellService } from 'cell-service';
import { staticView } from './static-view';
import { h } from './dom-util';
import view from './calendar.html';

@staticView({
  template: h(
    'template',
    { 'click.delegate': 'click()' },
    'Cell ${message}'
  ),
  resources: []
})
class Cell {
  count = 1;
  message = `Cell ${this.count}`;

  click() {
    this.message = `Cell ${++this.count}`;
  }
}

@staticView({
  template: view,
  resources: [Cell],
  imports: () => import('./test-cell')
})
export class Calendar {

  static inject = [CellService];

  constructor(service) {
    this.service = service;
  }

  load() {
    this.isLoaded = true
  }

  searchAll() {
    this.service.searchAllCells();
  }
}
```

Differences compared to dynamic loading:

* `@staticView` doesn't use `Loader` to load resources, as everything is declared statically, and thus, bundler friendly.
* `@staticView` can use local class as custom elements, custom attributes, unlike dynamic module loading, where classes need to be exported to be discovered. (As shown in the example above, `Cell` class is declared locally, and only used within `Calendar` view)
* `@staticView` can take pre-built `HTMLTemplateElement` as its template config, thus help avoid issue with IE Html parser, where it strip invalid markup, (e.g: `<template/>` element inside a `<table/>` element). This also helps user to use hyper script syntax to pre built the template, and theoretically JSX too (for fun purposes I guess)
* `@staticView` has config property `imports`, that return a promise, which can be used to declared code split point via `() => import('my-module')` . This is shown via `imports: () => import('./test-cell')` in above example, though you need to go to the demo in following link to check

### Further improvement:

* `aurelia-dialog` already accepts a class as ViewModel in its config, so it doesn't need any tweaks. That said, we can provide an alternative version, where we only accepts classes as view model, and remove the usage of dynamically loaded view model via module string, to help tree shaking

* `aurelia-router` should follow `aurelia-dialog` and accept a class directly at `moduleId` config

With the two points above checked, we can build statically analyzable Aurelia apps.

@EisenbergEffect @jdanyow @jods4 
@jsobell

### TODO

- [ ] test
- [ ] improve API
- [ ] profit ?

### DIFF
--- a/src/view-strategy.js
+++ b/src/view-strategy.js
@@ -253,3 +253,105 @@ export class InlineViewStrategy {
     return viewEngine.loadViewFactory(entry, compileInstruction, loadContext, target);
   }
 }
+
+@viewStrategy()
+export class StaticViewStrategy {
+  constructor(config) {
+    this.template = config.template;
+    this.resources = config.resources;
+    this.imports = config.imports;
+    this.moduleId = config.moduleId || '';
+    this.factoryIsReady = false;
+  }
+
+  /**
+  * Loads a view factory.
+  * @param viewEngine The view engine to use during the load process.
+  * @param compileInstruction Additional instructions to use during compilation of the view.
+  * @param loadContext The loading context used for loading all resources and dependencies.
+  * @param target A class from which to extract metadata of additional resources to load.
+  * @return A promise for the view factory that is produced by this strategy.
+  */
+  loadViewFactory(viewEngine, compileInstruction, loadContext, target) {
+    if (this.factoryIsReady) {
+      return Promise.resolve(this.factory);
+    }
+    let importResources;
+    let promise;
+    if (typeof this.imports === 'function') {
+      promise = Promise.resolve(this.imports());
+    }
+    else {
+      promise = Promise.resolve({});
+    }
+    return promise.then(importedResources => {
+      const imported = Object.keys(importedResources).map(i => importedResources[i]);
+      const resources = this.resources.concat(imported);
+      let template;
+      const container = viewEngine.container;
+      const appResources = viewEngine.appResources;
+      const viewCompiler = viewEngine.viewCompiler;
+      const viewResources = new ViewResources(appResources);
+      let resource;
+      let resourceTypeMeta;
+      // entry = this.entry = new TemplateRegistryEntry(this.moduleId);
+      template = /*entry.template =*/ typeof this.template === 'string' ? DOM.createTemplateFromMarkup(this.template) : this.template;
+      if (template.tagName !== 'TEMPLATE') {
+          throw new Error('Custom element template should be wrapped inside a <template> element.');
+      }
+      const requires = template.querySelectorAll('require');
+      if (requires.length > 0) {
+          for (var i = 0, ii = requires.length; ii > i; ++i) {
+          }
+          console.log("<require /> not supported in Static View Strategy yet");
+      }
+      const elDeps = [];
+      if (Array.isArray(resources)) {
+        for (var i = 0, ii = resources.length; ii > i; ++i) {
+          resource = resources[i];
+          resourceTypeMeta = metadata.get(metadata.resource, resource);
+          if (resourceTypeMeta) {
+            if (resourceTypeMeta instanceof HtmlBehaviorResource) {
+              if (resourceTypeMeta.attributeName === null && resourceTypeMeta.elementName === null) {
+                //no customeElement or customAttribute but behavior added by other metadata
+                HtmlBehaviorResource.convention(resource.name, resourceTypeMeta);
+              }
+              if (resourceTypeMeta.attributeName === null && resourceTypeMeta.elementName === null) {
+                //no convention and no customeElement or customAttribute but behavior added by other metadata
+                resourceTypeMeta.elementName = _hyphenate(resource.name);
+              }
+            }
+          } else {
+            if (resourceTypeMeta = HtmlBehaviorResource.convention(resource.name)) {
+              // either custom element / custom attribute
+              metadata.define(metadata.resource, resourceTypeMeta, resource);
+            } else if (resourceTypeMeta = ValueConverterResource.convention(resource.name)
+              || BindingBehaviorResource.convention(resource.name)
+              || ViewEngineHooksResource.convention(resource.name)) {
+              // something else that is not custom element / attribute
+              metadata.define(metadata.resource, resourceTypeMeta, resource);
+            } else {
+              // doesn't match any convention, and is an exported value => custom element
+              resourceTypeMeta = new HtmlBehaviorResource();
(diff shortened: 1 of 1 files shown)


Alexander-Taran:
But dynamic compose is not a viable case for this right?
Just to be clear

davismj:
Questions and comments:

This should also accept a string, right? `template: HTMLTemplateElement | string | null | undefined;`

> // address of the view, in case there is some dynamic loading involved

Dunno what that means without an example.

> @staticView can use local class as custom elements, custom attributes, unlike dynamic module loading, where classes need to be exported to be discovered. (As shown in the example above, Cell class is declared locally, and only used within Calendar view)

Why is that a benefit?

> @staticView can take pre-built HTMLTemplateElement as its template config, thus help avoid issue with IE Html parser, where it strip invalid markup, (e.g: <template/> element inside a <table/> element). 

Very cool.

> aurelia-router should follow aurelia-dialog and accept a class directly at moduleId config

Agreed!  It's something on my 2.0 radar from talks with @jods4. Additionally, router should accept a function for the moduleId that returns a class or a string for code splitting. Is there an issue for this already that you know of? Otherwise I'll create and we can associate it here.

bigopon:
@Alexander-Taran 
> But dynamic compose is not a viable case for this right?

`<compose/>`  config `viewModel`, works in the same way with `aurelia-dialog`, so we can still apply this, except that instead of string for module Id, it will be the custom element class:

```html
<template>
  <compose view-model.bind='vm' ></compose>
</template>
```

```js
export class MyElement {
  viewModels = {
    date: DateEditor,
    number: NumberEditor
  }

  // change this to another class if needed
  vm = this.viewModels.date;
}
```

> This should also accept a string, right? template: HTMLTemplateElement | string | null | undefined;

Yes, it accepts string, `HTMLTemplateElement` for normal view, `null / undefined` for `@noView`

> Why is that a benefit?

It's easier to scope local resources used in a view, thus aid app structuring / prototyping

> Agreed! It's something on my 2.0 radar from talks with @jods4. Additionally, router should accept a function for the moduleId that returns a class or a string for code splitting. Is there an issue for this already that you know of? Otherwise I'll create and we can associate it here.

you formatted the md wrong 😄  @davismj

bigopon:
> Agreed! It's something on my 2.0 radar from talks with @jods4. Additionally, router should accept a function for the moduleId that returns a class or a string for code splitting. Is there an issue for this already that you know of? Otherwise I'll create and we can associate it here.

I forgot to answer your question. I don't think there would be any issue, consider the fact that router only care about the first custom element it hits in the module anyway, which means giving it a class or a module should be no different. But I haven't dug into the router code enough to confirm this.

StrahilKazlachev:
@bigopon 
1. No `<compose>` and `aurelia-dialog` do not accept the same config options.
    - `<compose>` - `viewModel: string | object` - so either a *moduleId* or an object to be used as the VM.
    - `aurelia-dialog` - accepting *and a* `class` is possible since the `DialogService` uses `Origin.get` to convert it to `moduleId: string`. This part is causing the most issues with webpack, ask @jods4.
    - both are built on top of the `CompositionEngine`.
2. In `aurelia-router` the `RouteLoader` is an *abstract* class that must be provided for the router. The one provided by the `templating-router` is also on top of the `CompositionEngine`.
3. The dialog APIs *may* seem ready for supporting this case, but the internals are all based on dynamic composition.

bigopon:
With the help of aurelia/framework#858, probably we can make it easier by allowing config to opt out the default resources that aren't used, then we can make `<compose/>` v2, targeting the same name to support static scenario. Something like this

```js
typeof NO_AU_COMPOSE_V1 === 'undefined' ? Compose : ComposeV2
```

obedm503:
Really cool. Why still take a moduleId? This would mean Aurelia still needs a loader specific to the bundler used. Once this is introduced I doubt there would be a need for using moduleIds, or is there?

bigopon:
> Really cool. Why still take a moduleId? This would mean Aurelia still needs a loader specific to the bundler used. Once this is introduced I doubt there would be a need for using moduleIds, or is there?

I was thinking of supporting `<require/>` inside the markup, which will be converted to pure import. If it's hard to do, we can just scrap it. And no, module Id has no association with loader, it's acts as a base url.

bigopon:
On a second thought, a decorator seems to be an overkill, probably we can make it cleaner, by enabling convention via static property on the class:

```js
import view from './calendar.html'

class Calendar {

  static view = {
    template: view,
    resources: [],
    imports: () => import('splitted-bundle.js')
  }

  static inject = [CellService];

  constructor(service) {
    this.service = service;
  }

  load() {
    this.isLoaded = true
  }

  searchAll() {
    this.service.searchAllCells();
  }
}
```

obedm503:
A decorator seems better because it aligns itself with the rest of the API. What makes it seem like an overkill?

StrahilKazlachev:
We can have both, like with `inject`.

EisenbergEffect:
Instead of it being `@staticView`, I think we should just call it `@view`. Also, what about renaming `resources` to `dependencies`. I'm thinking of making that name change in vNext (since it's all container-based) and was wondering if this new API could be forward compatible. Finally, is there any way to combine the `imports` property with `dependencies`?

bigopon:
Renaming it to `@view` would be very nice, as if needed, we can make it behave like `@inject` and `static inject = [...]`

about `dependencies/imports`, probably we can unify everything into 1:

```js
@view({
  template: template,
  dependencies: () => [...] // let devs decide what to go in here, more flexibility
})
class myclass {

}
```

EisenbergEffect:
Sounds good. Let's go for that and see how it turns out.

obedm503:
can `dependencies` just be an array also? when it comes to local dependencies, most of the time they're not going to be dynamically imported

bigopon:
yeah, we could have that. final API could look like

```ts
interface IViewConfig {
  template: string | HTMLTemplateElement | Promise<string | HTMLTemplateElement>;
  dependencies?: Function[] | { (): Function[] | Promise<Function[]> };
}
```

EisenbergEffect:
@bigopon I've merged #614. Can we close this now?

Issue #609, opened 2018-03-27
Empty object injected as the Element for custom attributes


jack4it:
**I'm submitting a bug report**

* **Library Version:**
    "aurelia-animator-css": "^1.0.4",
    "aurelia-binding": "^1.7.1",
    "aurelia-bootstrapper": "^2.2.0",
    "aurelia-cookie": "^1.0.10",
    "aurelia-event-aggregator": "^1.0.1",
    "aurelia-framework": "^1.1.5",
    "aurelia-history": "^1.1.0",
    "aurelia-history-browser": "^1.1.1",
    "aurelia-http-client": "^1.2.1",
    "aurelia-router": "1.5.0",
    "aurelia-validation": "^1.1.3",

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
9.8.0

* **NPM Version:**
5.8.0

* **Yarn Version:**
1.5.1

* **JSPM OR Webpack AND Version**
webpack 4

* **Browser:**
Chrome 65

* **Language:**
TypeScript 2.7.2

**Current behavior:**
All custom attributes suddenly stop working after upgrading to all latest versions of the Aurelia packages along with the webpack 4 upgrade. Debugging shows that the injected Element for the attributes instances is an empty object.

**Expected/desired behavior:**
The injected Element should be the valid DOM element like it has always been. Now all my custom attributes are broken

**NOTE: I'm not using the CLI. I'm building with webpack via a node script**

StrahilKazlachev:
Just a sanity check, can you delete any `package-lock.json` file and do a clean install?

jack4it:
I'm actually using yarn and have tried to delete the yarn.lock file a couple of times and no luck... but, let me try npm install

jack4it:
Just tried `npm install` with no lock file in the first place. Exactly the same broken behavior. This happens to multiple different custom attributes in my codes.

StrahilKazlachev:
Then can you update the issue description by substituting the **NPM** section with **Yarn**, with the appropriate version.
Also can you specify whether your build is using the CLI?

jack4it:
Just updated the issue according to your comment

Alexander-Taran:
Usually it is a forgotten @autoinject
can you create a gist of one of custom elements?
https://gist.run/?id=8ae06845f2bdf4d533b072c00da82614

jack4it:
[...]

PR #610, opened 2018-03-28
feat(ViewResources): auto register

bigopon:
@EisenbergEffect


Alexander-Taran:
painful to read all those space removals (-:

src/view-resources.js
```diff
@@ -283,4 +288,41 @@ export class ViewResources {
...
+   * @param {Function} impl
+   * @returns {HtmlBehaviorResource | ValueConverterResource | BindingBehaviorResource | ViewEngineHooksResource}
+   */
+  autoRegister(container, impl) {
+    let resourceTypeMeta = metadata.get(metadata.resource, impl);
+    if (resourceTypeMeta) {
+      if (resourceTypeMeta instanceof HtmlBehaviorResource) {
+        if (resourceTypeMeta.attributeName === null && resourceTypeMeta.elementName === null) {
+          //no customeElement or customAttribute but behavior added by other metadata
+          HtmlBehaviorResource.convention(impl.name, resourceTypeMeta);
```
EisenbergEffect: The only real issue I see if the use of `Function#name` which will not be preserved when minifying code. Is the assumption for this feature that developers will always explicitly provide the decorator with the name so that it never falls back to the `Function#name`?

bigopon: Probably that's the expected behavior, since devs need to be aware of their script minimization, I think. And it's for `UglifyJs`, it's the `keep_classname`, that will do the job. Besides that, I see no harm on enforcing it if devs want to go all in on static code analysis

Alexander-Taran: >since devs need to be aware of their script minimization
nobody is aware of nothing (-:
example: https://github.com/aurelia/templating/issues/609#event-1547295242

bigopon: I am aware of that

bigopon: Probably we should treat this as private API first and only used by the config Api like framework configuration or static view. If folks want it, we could open it but view resources is something i would normally touch. We may also need a flag on the typing to tell it is for internal use. About doc, it would be better explained by a real usage of this api like config

EisenbergEffect: That makes sense. You can just add a doc comment that it's an internal API, not supported for public use and may change without warning, etc.

EisenbergEffect:
@bigopon I've merged #614. Can we close this now?

PR #611, opened 2018-04-05
feat(CompositionEngine): allow constructors

bigopon:


bigopon:
@EisenbergEffect This is the start of 2nd batch, I'll consolidate the logic a bit more and add some more tests, basically makes the composition behavior similar in both situation (string, and constructor). At the moment they are a bit different.

bigopon:
@EisenbergEffect It's ready for review

PR #614, opened 2018-04-14
feat(resource): static convention

bigopon:
This PR is a merge of

* #610 : auto resource registration
* #606 : static view strategy

together with 

* ability to specify aurelia resource type via static `resource` field, method on class, as a replacement for 
 `@customElement`, `@customAttribute`, `@valueConverter`, `@bindingBehavior`, `@viewEngineHooks` calls

usage examples:

#### Declaring a custom element `<normal-button/>`, `containerless`, with a bindable `type` attribute

```jsx

  class NormalButton {
    static resource () {
      return {
        name: 'normal-button',
        containerless: true,
        bindables: ['type']
      };
    }
    static view () {
      return '<template><button type.bind="type" >Normal button</button></template>';
    }

    // or jsx
    static view() {
      return <template>
        <button type$='type' onClick='shout()'>Normal button</button>
      </template>
    }
  }
```

no type specified = custom element

#### declaring a custom element '<fancy-button/>`

```jsx
  class NotSoFancyButton {
    static resource () {
      return {
        name: 'fancy-button',
        bindables: ['type']
      };
    }
    static view () {
      return '<template><button type.bind="type" >Fancy button</button></template>';
    }

    // or jsx
    static view() {
      return <template>
        <button type$='type' onClick='shoutInFancyWay()'>Fancy button</button>
      </template>
    }
  }
```

#### declaring a custom attribute `[date-picker]`

```jsx
class DatePicker {

  static inject () {
    return [Element];
  }

  static resource = {
    type: 'attribute',
    name: 'date-picker',
    bindables: [
      { name: 'date', defaultBindingMode: 'twoWay' }
    ]
  }

  constructor (element) {
    this.value = '';
    this.element = element;
  }

  attached () {
    this.element.type = 'date';
    this.element.addEventListener('input', this);
  }

  detached () {
    this.element.removeEventListener('input', this);
    this.element.type = 'text';
  }

  handleEvent (e) {
    this.value = this.element.value;
  }
}
```

#### Declaring an app custom element

```jsx
export class App {
  // string mean it's a custom element, name equals to the string
  static resource = 'app';

  static view() {
    return {
      template: `<template>
        <form>
          <input date-picker.bind='date' />
          <fancy-button></fancy-button>
        </form>
      </template>`,
      dependencies: [FancyButton, DatePicker]
    }
  }
    
  // or jsx
  static view() {
    return <template>
      <form>
        <input datePicker$='date' />
        <fancy-button type$='type'></fancy-button>
      </form>
    </template>
  }
}
```

The static `resource` will be looked for before traditional convention, such as export name.

### pros:
- helps devs write clean class, without any trace of aurelia, as demonstrated via above examples
- provides convention that easily works without transpiler, means later, it will be a simple `<script src="aurelia.umd.js"></script>` and plain javascript classes

### cons
- traditional convention via loader not applicable. (though there is probably a way to support loader, with a requirement that everything must be an absolute url)
- uncertain what will come

todo:

- [x] test
...

tobe discussed:

#### static `resource` signature

```ts
  export class ViewResources implements IViewResources {

    static convention(target: Function & { resource?: string | IResourceConfig | { (): string | IResourceConfig } })
    // ...
  }

  export interface IResourceConfig {
    type?: 'element' | 'attribute' | 'valueConverter' | 'bindingBehavior' | 'viewEngineHooks';
    name?: string;
    bindables?: Record<string, boolean | IBindablePropertyConfig>;
  }
```

  Not sure if `bindables` should be an object with each property map to a bindable property, or an array of either string or a bindable property config with name in it. With array, name is required if it's not a string

#### static field / method name

`resource` was picked because it aligns with internal term of Aurelia, but doesn't look friendly on famework user side. Probably `metadata` could do a better job. Should it be `metadata` instead of `resource` ?

@EisenbergEffect @jdanyow @jods4 

-----

Update 1:

- Features map

1. This PR, for easy resources recognition convention, resources registration via classes, static view for local view resources dependencies declaration via classes
    1.1 global resources declaration via classes https://github.com/aurelia/framework/pull/858
        --- templating-router resources https://github.com/aurelia/templating-router/pull/73
        --- templating-resources https://github.com/aurelia/templating-resources/pull/340
    1.2 Future pr, where `plugin` accepts a configure function, instead of module id, to be called in sequence when `aurelia.start` is called
2. Add new feature to composition engine: ability to compose if viewModel specified is a constructor https://github.com/aurelia/templating/pull/611
    2.1 Ability to use class as first parameter for `Aurelia.setRoot` https://github.com/aurelia/framework/pull/876
    2.2 Ability to use class / normal module for router https://github.com/aurelia/templating-router/pull/75


fkleuver:
I really like this idea!

- I think `bindables` (if multiple) should be an array. Having array/function convention on static properties is in line with how `static inject` works. And you could normalize a single object to an array for the occasional user that only wants to define a single one (often the case with attributes).

- Nothing wrong with the name `resource` imo. We've had a `resources` folder and `.globalResources()` call since day one, which always pertained to elements/attributes etc (exactly the thing you're addressing). Metadata on the other hand is a fairly abstract term to people who never worked directly with it (which is probably most people).

Just my 2c.

As for the loader thing, it might be interesting to add a `.static` extension loader plugin of sorts. Just a wild idea.

Now you can already inline html on a typescript class via `inlineView` etc, but that html is only accessible to that class. With your idea and loader integration, html defined on static properties of classes could  be shared with other classes and even directly referenced like `some-resource.static` (which loads `some-resource.ts` and gets the static `resource` property from it).

That, to me, seems **very** interesting because it would enable code-only projects. That can be really handy in some cases for plugins, and also be a big relief to users of unusual build systems who have trouble getting html resources properly loaded. It just sounds like a nice alternative to have.

Regardless if that will work (and happen), I'm really excited!

zewa666:
@bigopon could you update above samples with one using JSX for the view?

fkleuver:
@bigopon Will this only work when the properties are functions (seems that way right now)?

EisenbergEffect:
@bigopon I'd love to support both static methods and properties if possible.

bigopon:
No it can be either object or string or function resolving to one of those. Static method is what works today without transpiler so i used it

huochunpeng:
I can see the good will of supporting jsx as inline template, but the side effect is some unnecessary learning curve.

Now for users want to use inline template, they need to understand how jsx works, and also how to map its features to Aurelia. We don't need that learning curve.

Instead of:
```jsx
  static view () {
    return <template>
      <input datePicker$='date' />
      <button type$="type" onClick='shout()'>Normal button</button>
      {'${message}'}
    </template>;
  }
```

We can just write:
```js
  static view () {
    return `<template>
      <input date-picker="date">
      <button type="type" click.trigger="shout()">Normal button</button>
      \${message}
    </template>`;
  }
```

There are two minor issues around using es6 multiline string literal.
1. there is no syntax check for inline html from editor/transpiler.
2. need to use `\${m}` to bypass es6 string interpolation.

To make it better, but instead of using babel plugins `syntax-jsx`, `transform-react-jsx`, I suggest to develop new plugins `syntax-jsau`, `transform-jsau` to support inline html.

> name `jsau` might be miss-leading, as this new format has no dependency of aurelia. It just tries to support inline html (real html) in js file.

This is actually technically simpler than trying to mock `React.createElement`.

* `syntax-jsau` is easy to implement than `syntax-jsx`, because we can use a DOM parser to parse html, I mean real html by standard, not jsx html.
* `transform-jsau` is really dumb, just wrap the contents in a js string, instead of parsing it into `React.createE ...`.
* need some work on `babel-eslint` to support `jsau` syntax check.
* need more work on popular editors to support `jsau` syntax check.

So we can finally write:
```js
  static view () {
    return <template>
      <input date-picker="date">
      <button type="type" click.trigger="shout()">Normal button</button>
      ${message}
    </template>;
  }
```

zewa666:
@huochunpeng jsx was purely out of interest to see whether it's doable. It would also provide a nicer migration for existing React code. So I definitely don't envision to push users into JSX direction, more just to know that it kinda works

huochunpeng:
That makes sense. I can see from stackblitz jsx is implemented purely by a mock of React.createElement.

bigopon:
A custom plugin to display html in jsx is awesome, but would be a lot of work I believe. I have no experience of that, maybe @eriklieben, who has been working on vscode extension, can give some comments. Personally I believe complex markup needs to be in their own well formatted file. So final usage would be a plugin to inline the template during build time (single file module / component etc), or devs explicitly import view in the view model file and use it. Though inline view with template literal is very good for demos / getting started. Also another way is to get it directly from `index.html`, as view works with `HTMLTemplateElement` too:

```html
<!-- index.html -->

<body>
  <div id="app-main-template">
    // heavy markup prerender by server
  </div>
</body>
```

```js
class App {
  static view() {
    let template = document.createElement('template');
    template.content.appendChild(document.querySelector('#app-main-template'));
    return template;
  }
  // ...
}
```

bigopon:
@EisenbergEffect Tests for all new features added and ready for review.

huochunpeng:
> So final usage would be a plugin to inline the template during build time

As long as not using webpack, this can be a small gulp plugin to find matching vinyl files (js and html) and merge them, at gulp stream flush stage.

huochunpeng:
Had a brief talk with @bigopon about sub-class support, glad to know it all works.

Here is some code example for reference.

```js
class AnotherButton extends NormalButton {
  static resource() {
    let res = super.resource();
    res.name = 'another-button';
    return res;
  }

  shout() {
    super.shout();
    alert('Actually clicked another button :-)');
  }
}
```

EisenbergEffect:
@bigopon Since things have changed a bit in the latest set of iterations, can you update this issue with the full list of PRs, the order they need to go in and any other relevant information? I can start with "ring 0" this week and we can release them all over the next 2-3 weeks, while we work on some documentation.

bigopon:
@EisenbergEffect i have updated the PR with the road map how things go together

EisenbergEffect:
Thanks!

obedm503:
does this pr still work with decorators like #606 did?

bigopon:
@obedm503 New check was added for `ViewLocator` to check static `view` property on the class, not much else. So there is no decorator. But it can be easily created:

```js
export function view(config: string | IStaticViewConfig): any {
  return function deco(target: Function) {
    target.view = config;
  }
}
```

@EisenbergEffect should be provide a decorator ?

obedm503:
It'd be nice to have. I prefer decorators to static properties

fkleuver:
Probably good idea to include that decorator source code in the docs as well. It's nice to define your own decorators in larger apps as a way to apply common logic/conventions and stuff. More people should know about it.

src/view-locator.js
```diff
@@ -50,6 +50,18 @@ export class ViewLocator {
       value = value.constructor;
     }
 
+    // static view strategy
+    if (value.view) {
```
jods4: This is a breaking change.
Use a symbol instead?

src/view-resources.js
```diff
@@ -67,6 +73,72 @@ interface ViewEngineHooks {
...
 */
 export class ViewResources {
+
+  static convention(target: Function) {
+    let config = target.resource;
+    let resource;
+    if (config) {
+      if (typeof config === 'string') {
+        // it's a custom element, with name is the resource variable
+        // static resource = 'my-element'
```
jods4: custom attribute being quite useful (in my opinion), is it worth it to have a shortcut for that as well?
For example:
`"data-grid"` -> custom element named `data-grid`
`"@blink"` -> custom attribute named `blink`

src/view-resources.js
```diff
@@ -67,6 +73,72 @@ interface ViewEngineHooks {
...
+        resource = new HtmlBehaviorResource();
+        resource.elementName = config;
+      } else {
+        if (typeof config === 'function') {
+          // static resource() {  }
+          config = config();
+        }
+        if (typeof config === 'string') {
```
jods4: Why is `resource() { return "my-element" }` a different code path than `resource: "my-element"`.

It's not even duplicated, it's actually different code. Shouldn't it be the same behaviour?

src/view-resources.js
```diff
@@ -67,6 +73,72 @@ interface ViewEngineHooks {
...
+          // static resource() { return 'my-custom-element-name' }
+          config = { name: config };
+        }
+        // after normalization, copy to another obj
+        config = Object.assign({}, config);
```
jods4: Nitpick: could be a `else` from previous checks, avoids one allocation and one copy.

Also: this could crash. For example if I do `resource: 4` in my code or any type that's neither a string, an object or a function returning `string | object`. `resource() { return null }` would equally crash, etc.

src/view-resources.js
```diff
@@ -67,6 +73,72 @@ interface ViewEngineHooks {
...
+        let resourceType = config.type || 'element';
+        switch (resourceType) {
+          case 'element': case 'attribute':
+            resource = new HtmlBehaviorResource();
+            if (resourceType === 'element') {
+              resource.elementName = _hyphenate(config.name || target.name);
+            } else {
+              resource.attributeName = _hyphenate(config.name || target.name);
+            }
+            if (config.templateController) {
```
jods4: Nit: I feel like `in` would be more consistent with `defaultBindingMode` check just below.
It would also avoid me thinking about whether ignoring `templateController: false` is totally ok or has tiny chance of screwing us up.

src/view-resources.js
```diff
@@ -67,6 +73,72 @@ interface ViewEngineHooks {
...
+            // one time default binding mode is 0, cannot check for falsy value ... T_T
+            if ('defaultBindingMode' in config) {
+              config.attributeDefaultBindingMode = config.defaultBindingMode;
+              delete config.defaultBindingMode;
+            }
+            // just copy over. Devs are responsible for what they specify in the config. good luck!.
+            Object.assign(resource, config);
+            break;
+          case 'valueConverter':
+            resource = new ValueConverterResource(camelCase(config.name || target.name)); break;
```
jods4: Nit: same line `break` is hard to see.
Same on the next two `case`.

src/view-resources.js
```diff
@@ -67,6 +73,72 @@ interface ViewEngineHooks {
...
+            resource = new ViewEngineHooksResource(); break;
+        }
+      }
+      // check for bindable registration
+      let bindables = typeof config === 'string' ? undefined : config.bindables;
+      if (Array.isArray(bindables) && resource instanceof HtmlBehaviorResource) {
+        for (let i = 0, ii = bindables.length; ii > i; ++i) {
+          let prop = bindables[i];
+          if (!prop || (typeof prop !== 'string' && !prop.name)) {
+            throw new Error('Invalid bindable property. Expected either a string or an object with "name" property.');
```
jods4: It would be very helpful for users to have context in errors.
For instance the name of class being processed `target.name` (during dev it wouldn't be obfuscated / minified away) and maybe the index `i` of faulty bindable.

src/view-strategy.js
```diff
@@ -253,3 +254,77 @@ export class InlineViewStrategy {
     return viewEngine.loadViewFactory(entry, compileInstruction, loadContext, target);
   }
 }
+
+@viewStrategy()
+export class StaticViewStrategy {
+
+  template: string | HTMLTemplateElement;
+  dependencies: Function[] | { (): Function[] | (Function | Promise<Function[] | Record<string, Function>>)[] };
```
jods4: I _think_ I get this based on previous discussions we had.
But it really needs some explanations and some tidy-up (notice for example `Function[]` is redundant in your return type).

If I get this right, you 100% need to explain that:
- it returns an array of promises rather than a promise for an array to easily support `[() => import("x"), () => import("y")] pattern`.
- that the `Record<string, Function>` is meant for lazy loading based on resource name, so the `Function` here is a factory unlike all other `Function` in that declaration, which are class ctor.

src/view-strategy.js
```diff
@@ -253,3 +254,77 @@ export class InlineViewStrategy {
...
+  * @param loadContext The loading context used for loading all resources and dependencies.
+  * @param target A class from which to extract metadata of additional resources to load.
+  * @return A promise for the view factory that is produced by this strategy.
+  */
+  loadViewFactory(viewEngine: ViewEngine, compileInstruction: IViewCompileInstruction, loadContext: IResourceLoadContext, target: any): Promise<IViewFactory> {
+    if (this.factoryIsReady) {
+      return Promise.resolve(this.factory);
+    }
+    let deps = this.dependencies;
+    deps = typeof deps === 'function' ? deps() : (deps ? Array.isArray(deps) ? deps : [deps] : []);
```
jods4: Nit: no, thanks.
You did get the precedence of everything right, but that's too much trivia for real code.
I suggest _at least_ reordering the checks so that there is no ambiguity in the ternary operators results:
`isFunction ? deps() : isArray ? deps : deps || []`
That's still a bit cryptic to me, but at least there's no ambiguity in results. 
Note that it works because `isArray` simply returns false for anything that's not an array, including `null`, `undefined`.

src/view-strategy.js
```diff
@@ -253,3 +254,77 @@ export class InlineViewStrategy {
...
+        // when composing without a view mode, but view specified, target will be undefined
+        viewResources.autoRegister(container, target);
+      }
+
+      for (let dep of dependencies) {
+        if (typeof dep === 'function') {
+          resource = viewResources.autoRegister(container, dep);
+        } else if (dep && typeof dep === 'object') {
+          for (let key in dep) {
+            resource = viewResources.autoRegister(container, dep[key]);
```
jods4: So... I don't quite get it.
What's the point in supporting a record if we do nothing with `key`?
Is it so that future versions could lazy-load resources when their name is first encountered in a view?

That's a cool future feature, but then I think the best choice is to not accept a record until then. 
Because lazy loading is a breaking change, so we don't want people to use this syntax until we actually implement the feature.

Even so, it's not implemented correctly. As lazy-loading implies a factory, a correct eager-loading implementation would be: `dep[key]()`.

jods4: Oh wait, I think I get it now... 
This is for the case `() => import("y")`, isn't it?
The record is actually an ES module?

Makes sense then. 
I like how we try to find all resources in said module, rather than just the first one as in older Aurelia code 👍 
No reliance on `default`, then.

src/view-strategy.js
jods4: We might want a factory returning a single resource, i.e. a function?
It might be a typo but it's not in the typedef above.

There are better use cases, but this: `() => [ async () => (await import("x")).MyElement ]`.

src/view-locator.js
bigopon: You mean a static `view` method / field on a class detection is a breaking change ?

The reason it went with `static resource / view` instead of symbol is that devs can easily create a custom element, without importing anything from Aurelia. I'm not sure about symbol here. Maybe @EisenbergEffect wants to discuss too.

src/view-resources.js
bigopon: the `@` seems nice, but also out of nowhere I think, maybe we can use existing definition: `[custom-attribute]` like css selector ?

src/view-resources.js
bigopon: That would be too passive, probably should be job for documentation, or I think we can throw an error to let devs know ?

bigopon: We could throw a custom error so devs could know easier what going on in case.

src/view-resources.js
bigopon: I missed this. Will add a fix

src/view-resources.js
```diff
@@ -67,6 +73,72 @@ interface ViewEngineHooks {
...
+
+  static convention(target: Function) {
+    let config = target.resource;
+    let resource;
+    if (config) {
+      if (typeof config === 'string') {
+        // it's a custom element, with name is the resource variable
+        // static resource = 'my-element'
+        resource = new HtmlBehaviorResource();
+        resource.elementName = config;
```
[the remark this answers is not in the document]
bigopon: 🐛 😱

src/view-resources.js
```diff
@@ -283,4 +355,45 @@ export class ViewResources {
...
+   */
+  autoRegister(container, impl) {
+    let resourceTypeMeta = metadata.get(metadata.resource, impl);
+    if (resourceTypeMeta) {
+      if (resourceTypeMeta instanceof HtmlBehaviorResource) {
+        if (resourceTypeMeta.attributeName === null && resourceTypeMeta.elementName === null) {
+          //no customeElement or customAttribute but behavior added by other metadata
+          HtmlBehaviorResource.convention(impl.name, resourceTypeMeta);
+        }
+        if (resourceTypeMeta.attributeName === null && resourceTypeMeta.elementName === null) {
```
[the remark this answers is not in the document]
bigopon: It's the way we check for custom element and custom attribute.

Try with convention first, if nothing matches > fallback to element

src/view-strategy.js
bigopon: Yes, `Record<string, any>` is for `import('x')`. That said, I just realized it's doesn't handle the case where an export is not a function. Probably we should ignore any export value that is not a function.

src/view-strategy.js
bigopon: I think I have fixed this in the followup PR. Reading it again does feel terrible

src/view-resources.js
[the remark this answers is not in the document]
bigopon: Rethinking about it, probably I will have to do it differently: if `name` was specified, use it and lower case it, otherwise hyphenate the class name. It's more consistent with current way how it works.

src/view-strategy.js
bigopon: Yes, It is also consistent with the way we currently load resources if it was string: every thing will be registered. And sure, docs for this work gonna be fun...

src/view-resources.js
jods4: Yes, that would be more "native" to the web platform. Dunno where I got the `@` idea from. 😀

src/view-locator.js
jods4: I mean that existing code that contains classes with a `view` static field will trigger this code with unintended effects.

Unless we are sure that existing code will not execute this (e.g. if you need to use new API entry points)?

src/view-resources.js
jods4: Yes. 
A random unexplained crash is bad. An error that says _"view of {target.name} should be a string or an object"_ is a lot more helpful.

src/view-resources.js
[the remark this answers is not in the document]
jods4: I agree. It's strange that we modify an explictely defined name.

Hypenate made perfect sense when we derived the element from class name.

src/view-resources.js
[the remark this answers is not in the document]
jods4: I understand what it does but it looks weird. Basically you do:
```
if (A) {}
if (A) {}
```
If the first `A` is false, the second has to be false as well and is redundant.
Nitpicking anyway, it works fine.

src/view-locator.js
StrahilKazlachev: I'm using `static view;` in the VMs I feed to my custom `<compose>`like elements - just a feedback for possible scenarios this could be found now.
I do like the `symbol` idea. We could use `Symbol.for()`, define and export the string keys in Aurelia. Now if you want to import the keys you can, or just type them yourself.

src/view-resources.js
bigopon: then do we fully embrace symbolic characters ? like `'|valueConverter'` and `'&bindingBehavior'` ? Maybe we can comeback to it later as an enhancement, otherwise it would be quite hard to explain in doc about the inconsistency.

src/view-locator.js
bigopon: @StrahilKazlachev That's a nice idea. We can do `Symbol.for('au.view')`, but then I think it kind of defeats the point of having a static method / field. Maybe view is too generic, lets make it `viewStrategy` ? @EisenbergEffect @jods4 ?

jods4: @bigopon is this really a regression or not? I mean: will existing applications go through this code, or do you have to use a new method to enable this?
From the rest of the changes, I have the impression that this has precedence over conventions, so existing classes _will_ go through this, right?

`viewStrategy` is less likely to be used than `view`... but it's more cumbersome for users and there's always a chance someone uses it already. Something like `au-view` is both shorter and _very_ unlikely to be used (if only because of the dash!).

`Symbol` is made to avoid clashes. But it's even less convenient to use:
```ts
class MyElement {
  [Symbol.for("au-view")]() { return "<div>Hello</div>"; }
}

// Alternative
import { view } from "aurelia-symbols";

class MyElement {
  [view]() { return "<div>Hello</div>"; }
}
```

That said when it comes to convenience I would probably prefer a decorator and then the key/symbol doesn't matter.

src/view-resources.js
jods4: Yes, that was an enhancement suggestion. 
We can absolutely come back to it later when we have more feedback on the v1.
`[my-attribute]` is not a valid element name anyway, so it won't be a breaking change to add that kind of support later... if we want to.

bigopon:
@EisenbergEffect 

Summary from @jods4 & @StrahilKazlachev 's review

Checking for a static method / field `view` could introduce breakages into existing app that use this name.

---

I think we may need to replace it with symbol, e.g `Symbol.for('au-view')` instead of normal name. So that folks who wish to use static can still use it, though less convenient than old name:

```js
class MyElement {
  static [Symbol.for('au-view')] = '...'
  // or
  static [Symbol.for('au-view')]() {
    // ...
  }
}
```

Same thing may apply for `resource`. Besides that, there are few other issues related to convention that I will fix after we decide what to do with naming, like this https://github.com/aurelia/templating/pull/614#discussion_r183576073

bigopon:
That will work. Then how would you suggest for interop between static opt in and decorators ?

EisenbergEffect:
If someone uses a decorator that would create a static prop, but they haven't enabled the feature, we should throw with a clear error message explaining that they need to opt into the new model and how to do that.

bigopon:
There is probably no reason why using decorators only would not work, should throwing error would be unexpected ? Or you suggesting that we need a strong switch for mental model so folks can be aware of side effect of these static features ?

EisenbergEffect:
I was thinking that you thought there would be clash between the two, so I was recommending a potential solution. I may not have understood your question then...

bigopon:
Oh I see what you meant. Because decorators resolves to static field, that's why you suggest throwing an error. I was thinking decorator could generate some metadata, while static field ... declare its static bit and we make them interop-able. But we could do as you suggested, always resolve to static method / field and add a flag to opt it in.

EisenbergEffect:
Yep :)

jods4:
@EisenbergEffect I know I can't have it both ways: back-compat and easy use of new releases...

But still I'm not fan of the opt-in API. This way will become the way to go, especially with Webpack and co., but if it's behind a config it won't be obvious and easy to get started.

What if we use an inconvenient enough property, so that we're 90% sure it won't break anyone. I previously suggested:
```ts
class X {
  static ["au-view"]  = "<img>";
}
```
It's not that bad, but the presence of dash makes it _highly unlikely_ to have been used by someone.

jods4:
One more thing I just thought about: we _need_ this enabled by default, somehow.

Here's why:

This has tremendous value because it would enable us to drop our Webpack plugin in new projects and support many bundlers. This is the main motivation behind this work.

To drop the Webpack plugin, we need our core libs to be updated to use this new static conventions.

If we update core libs to use this, it's not an opt-in anymore.

bigopon:
I support what @jods4 suggested. Using a switch also gives plugin authors dilemmas if they wanna use this feature, it will be eventually on for the rest of the application if a plugin decides to turn it on. Unless it also accepts an additional static prop to turn it on and off per class

obedm503:
But that's on the way to too much config. I honestly don't see a problem with using symbols or even Reflect.metadata as long as it's managed carefully

fkleuver:
@obedm503 Main problem with symbols is they would require a lot more changes (and in some places plain additions, which can cause performance degradation). They aren't discoverable in the same way as normal properties. `for .. in`, `Object.keys()`, `Object.getOwnPropertyNames` and `Object.hasOwnProperty` don't reveal symbols.

However I don't see any issues with poor man's Reflect.metadata, which is to just put a `__aurelia__` property on everything with `Object.create(null)`. Would arguably make things even simpler and more performant than they are now.

obedm503:
Right, forgot about symbols not being enumerable

EisenbergEffect:
Why don't we use `$resourse` and `$view`? That would avoid conflicts and enable a simpler syntax for usage. There's also precedent with this type of naming pattern in other frameworks. Additionally, I'm using a similar naming scheme in the vNext work.

jods4:
I agree there's a precedent for `$` variables being "special" or "reserved", it's quite unlikely someone has a static method or field starting with `$` on a VM. 

I would even say that it hints the fact that the field is "magical", which is a rather good thing. Conventions can sometimes be hard to understand for newcomers.

So if there's no better proposal... `$view`?

fkleuver:
Agreed, `$` is an unambiguous prefix. If it's also used in vNext then that will make the transition that much smoother.

obedm503:
why not something super ugly and even less likely to be already used like `__aurelia__view__` but when using it aurelia could export an enum to make usage easier
```ts
// some aurelia file
export enum AureliaMagicFields {
  view = '__aurelia__view__'
}

// user - app.ts
export class App {
  static [AureliaFields.view] = []
}
```
I recommend this for a few reasons: 
- it's very unlikely that this static property will be reassigned later in code, so the inconvenience of a computed key is not that bad. generally it will just be declared up front
- it adds strength to the api. if someone misspells `$veiw` it will probably take longer to debug than misspelling `[AureliaMagicFields.veiw]` because either the linter or their compiler will complain

I guess it doesn't even have to be an exported enum, just a constant string will do as well. also, if aurelia ever wanted to move to symbols for these kinds of magic/conventional fields, the breakage, from the users' perspective, wouldn't be as bad if they're already using computed fields

jods4:
@obedm503 
If we go with something based on an import, like :
```ts
import { view } from "aurelia-symbols";

export class App {
  static [view] = "";
}
```
Then I would argue there is absolutely no reason not to go with a symbol instead of a string.

obedm503:
@jods4 I think it's best form every package to export their own constants instead of a single package with all of them. say this idea was accepted and user in other places, like the lifecycle methods

```ts
// import {viewConstants} from 'aurelia-templating';
enum viewConstants {
    view = '__au-view__', // or = Symbol('au-view')
    resource = '__au-resource__', // or = Symbol('au-resource')

    attached = '__au-attached__', // or = Symbol('au-attached')
    detached = '__au-detached__', // or = Symbol('au-detached')
}

// import {routerContants} from 'aurelia-router';
enum routerConstants {
    activate = '__au-activate__', // or = Symbol('au-activate')
    deactivate = '__au-deactivate__', // or = Symbol('au-deactivate')

    // previously `configureRouter`
    configure = '__au-configure__', // or = Symbol('au-configure')'
}

class Component {
    static [viewConstants.resource] = 'x-component';
    static [viewConstants.view] = '<template><button>Normal button</button></template>';

    handleClick = () => {}
    [viewConstants.attached]() {
        document.addEventListener('click', this.handleClick);
    }
    [viewConstants.detached]() {
        document.removeEventListener('click', this.handleClick);
    }
}

class Parent {
    router: any; // Router
    // previously `configureRouter`
    [routerConstants.configure](config, router) {
        this.router = router;
    }
}

class Page {
    static [viewConstants.view] = '<template><x-component></x-component></template>';

    [routerConstants.activate](params) {}
    [routerConstants.deactivate](params) {}
}
```
just for the record I'm not suggesting `viewConstants` or `routerConstants` as names, these are just for the example

@EisenbergEffect  @bigopon what do you think of the idea of using contants (strings or symbols) for aurelia-magic/reserved methods and properties?

if the idea is green-lit I could work on a pr to use constants for static properties and functions. I'm still not totally sure about lifecycle methods.

fkleuver:
I use symbols exclusively in a plugin where I wrap and intercept the `configureRouter` method of a user's view model. I think that's one of the few use cases where using a symbol is the only correct thing to do.

Most libraries use generated/long names when modifying globals or objects they do not own, such as html elements. Aurelia also does this in those cases.

When it's the user putting these properties on their *own* classes, why worry so much about potential conflicts? Users will have more stuff to learn about and more stuff to import. I feel like it's unnecessary bloat but maybe I'm missing something..

obedm503:
As I said, when it comes to lifecycle methods I'm not sure it's necessary. But when it comes to something like static view or resource, there IS a potential for conflict since these are new reserved names. It's entirely possible that someone's using one of these names already and by introducing this feature their code will break. 

The suggestion of using imported constants removes that possibility for breakage

PR #616, opened 2018-04-21
feat(decorators): add decorators for view and resource config

bigopon:
Follow up PR of #614 to have consistency with `@inject` and `static inject` 

I'm not certain about `@resource`, could be a breaking change but I don't think it's in use outside of templating-resources

@EisenbergEffect

### DIFF
--- a/dist/aurelia-templating.d.ts
+++ b/dist/aurelia-templating.d.ts
@@ -948,67 +948,12 @@ export declare class ShadowDOM {
   static distributeNodes(view?: any, nodes?: any, slots?: any, projectionSource?: any, index?: any, destinationOverride?: any): any;
 }
 
-export declare interface IStaticResourceConfig {
-  /**
-   * Resource type of this class, omit equals to custom element
-   */
-  type?: 'element' | 'attribute' | 'valueConverter' | 'bindingBehavior' | 'viewEngineHooks';
-  /**
-   * Name of this resource. Reccommended to explicitly set to works better with minifier
-   */
-  name?: string;
-  /**
-   * Used to tell if a custom attribute is a template controller
-   */
-  templateController?: boolean;
-  /**
-   * Used to set default binding mode of default custom attribute view model "value" property
-   */
-  defaultBindingMode?: bindingMode | keyof typeof bindingMode;
-  /**
-   * Flags a custom attribute has dynamic options
-   */
-  hasDynamicOptions?: boolean;
-  /**
-   * Flag if this custom element uses native shadow dom instead of emulation
-   */
-  usesShadowDOM?: boolean;
-  /**
-   * Options that will be used if the element is flagged with usesShadowDOM
-   */
-  shadowDOMOptions?: ShadowRootInit;
-  /**
-   * Flag a custom element as containerless. Which will remove their render target
-   */
-  containerless?: boolean;
-  /**
-   * Custom processing of the attributes on an element before the framework inspects them.
-   */
-  processAttributes?(viewCompiler: ViewCompiler, resources: ViewResources, node: Element, attributes: NamedNodeMap, elementInstruction: BehaviorInstruction): void;
-  /**
-   * Enables custom processing of the content that is places inside the custom element by its consumer.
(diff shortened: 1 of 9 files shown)


bigopon:
@EisenbergEffect i moved them to the JS file, next to its usage. Is that ok or should it be removed completely ?

EisenbergEffect:
@bigopon Have you run the lint/build to see if it actually works? 😁

jods4:
I like that!

Could you provide a short summary of the shape of added APIs? There was a lot of definitions moved so it's a bit hard to see in the diffs.

Of course main issue here is the ongoing discussion about backcompat. Adding a `resource` field to classes is maybe even more likely to break in someone's code than `view`.

Maybe I'm crazy but: would it be too much bloat to export a symbol, have the attributes use that. This would be always be available and it is backward compatible.
Plus having an opt-in for `view` static field?

So attributes (or direct symbol usage) would easily work for everyone, and the loader would fallback and check for simple strings only if you opted in?

bigopon:
If we decided to move forward with `au-view`, `au-resource` should be good. But here is what makes the most concerns, it creates inconsistent api. We have `@inject` and `static inject`, not the same way with resource and view @EisenbergEffect @jods4
---
@jods4 i will add it after some more commits.

bigopon:
@jods4 @EisenbergEffect last commit should fix few issues in #614 . 

changes:

* change `resource` and `view` to `$resources` and `$view`. 
    * name resolving in config: will always call `_hyphenate` for custom element / attribute (we are doing this in `@customElement`, `@customAttribute` so I think it should be consistent.
    * add warning when something is misconfigured
* add `@view` decorator, which basically add static `$view` field on class
* modify `@resource` decorator, which now accepts string / POJO and assign to static `$resource` field on class. Otherwise, keeps old behavior: define the metadata
* adjust tests

---

For interop (between static configure and metadata definition via decorator e.g `@bindable`) PR at #617 , it will have to wait until after this PR has settled. @jods4 Would be nice if you could give some feed back on the idea too

jods4:
@bigopon Which idea exactly? Do you want me to have a look at #617?

bigopon:
Yes.

I think I should recap the things as the features span over 2 PRs, making it annoying to handle.

#614 introduces

  1. a way to statically link custom element view model and its view together. The view data can be defined via object properties such as `template` for template, and `dependencies` for its dependencies. Dependencies should be classes. It does so via looking for a static method / field `$view` on a custom element class. This PR introduces a decorator to for folks who prefer decorator over static method / field.
  2. a way to define resources type without importing any metadata helper functions from Aurelia, such as `@customElement`, `@customAttribute`, `@templateController`, etc ..., It does so via looking for a static method / field `$resource` on a resources view model class.
  3. a way to conveniently initialize and then register a resource class, base on its metadata, with a `ViewResources` instance, without knowing exactly what resources type it should tell the `ViewResources` to register it to.

---
(1) and (3) are critical for a plugin-less Aurelia version, that is friendly to modern bundlers.

----

Because `(2)` introduces a completely different way to define resources type, which has been done via metadata decorators, it is now need some clarification around interop between them. #617 adds the ability to recognize metadata, either it was defined in static method / field `$resource` or via decorator like `@bindable`.

Basically `(2)` is nice to have, as it helps folks to have "Aurelialess" classes, which still be initialized correctly in an Aurelia application, like following example:

```js
export class DatePicker {
  static $resource() {
    return {
      type: 'attribute',
      bindables: [ { name: 'value', defaultBindingMode: 'twoWay' }, 'format']
    }
  }

  valueChanged(date) {

  }

  // ...
}
```

Theoretically, they are great for plugin, as it will be simple class exports for any plugin. For application, folks may mix between `bindables: any[]` in `$resources` like above example, together with `@bindable`. We can have it all or nothing, or we can have them interop-able.

@EisenbergEffect @jods4 @jdanyow

EisenbergEffect:
@bigopon This seems fine to me. I'm ready to merge it if you are.

bigopon:
@EisenbergEffect I think it's ready

PR #618, opened 2018-04-21
removed code sample which doesn't belong  to that paragraph

hariseldon78:


EisenbergEffect:
Thanks @hariseldon78 !

PR #630, opened 2018-06-27
Updated dependencies for aurelia-binding (removed 1.x dependency)

bulldetektor:
To be able to install via JSPM (which doesn't support comparator sets for versioning)


Alexander-Taran:
updated on master.
can close this

PR #631, opened 2018-07-08
fix(composition-engine): align class and moduleId behaviour

StrahilKazlachev:
I'm expecting regardless whether I use *moduleId* or *class* for `.viewModel` to get the same behaviour for instantiating the VM.
Currently if using a *class* and it has no registrations(@*transient*, etc.) or there is no resolver registered in the `.container`/`.childContainer`(from the `CompositionContext`) it will be instantiated and registered as singleton in the *root container*. In contrast when using *moduleId* the loaded *class* will be *auto registered* in the *child container*.
@bigopon is it intended?

### DIFF
--- a/src/composition-engine.js
+++ b/src/composition-engine.js
@@ -169,8 +169,8 @@ export class CompositionEngine {
 
     if (typeof context.viewModel === 'string') {
       context.viewModel = context.viewResources
-          ? context.viewResources.relativeToView(context.viewModel)
-          : context.viewModel;
+        ? context.viewResources.relativeToView(context.viewModel)
+        : context.viewModel;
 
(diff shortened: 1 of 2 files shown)


bigopon:
Thanks @StrahilKazlachev , nice catch. It is not intended, you can see what I used here https://github.com/StrahilKazlachev/templating/blob/5c60a572cbd0268e72aaf040ad373c12141e4c7d/src/composition-engine.js#L202

It should come from `childContainer` first. Definitely a bug. Could be also nice if you could add a test to ensure subsequent composition gets the right container @StrahilKazlachev

bigopon:
Lgtm <3

@EisenbergEffect

PR #637, opened 2018-08-30
Fixed a memory leak where unbind was not called if view was already detached

dannyBies:
This is a fix related to [this issue](https://github.com/aurelia/templating/issues/541#issuecomment-416160922)

We noticed that some of our unbind's on components where not being called which caused a memory leak in our application. This pull request intends to fix this.

I'm not a 100% sure if this is the correct place to fix this issue and if it causes any different behaviour than intented.

### DIFF
--- a/src/view-slot.js
+++ b/src/view-slot.js
@@ -484,13 +484,11 @@ export class ViewSlot {
     let children = this.children;
     let ii = children.length;
 
-    if (this.isAttached) {
-      for (let i = 0; i < ii; ++i) {
-        if (returnToCache) {
-          children[i].returnToCache();
-        } else {
-          children[i].detached();
-        }
+    for (let i = 0; i < ii; ++i) {
+      if (returnToCache) {
+        children[i].returnToCache();
+      } else if (this.isAttached) {
+        children[i].detached();
       }
     }
 
--- a/test/view-slot.spec.js
+++ b/test/view-slot.spec.js
@@ -281,6 +281,16 @@ describe('view-slot', () => {
         viewSlot.remove(view, true);
         expect(view.returnToCache).toHaveBeenCalled();
       });
+
+      it('calls return to cache even when detached is true', () => {
+        spyOn(view, 'returnToCache');
+
+        viewSlot.add(view);
+        viewSlot.detached();
+        viewSlot.removeAll(true);
+
+        expect(view.returnToCache).toHaveBeenCalled();
+      });
     });
 
     describe('.removeAll', () => {


bigopon:
@dannyBies Thanks for this. I just checked other `remove...` methods and they all handle `returnToCache` properly except `remove all`, that's why we have been having issues only with `Repeat`.
The fix I introduced before to `_projectionRemoveAll` wasn't sufficient:
```diff
     let children = this.children;
+    let ii = children.length;
     if (this.isAttached) {
-      for (let i = 0, ii = children.length; i < ii; ++i) {
-        children[i].detached();
+      for (let i = 0; i < ii; ++i) {
+        if (returnToCache) {
+          children[i].returnToCache();
+        } else {
+          children[i].detached();
+        }
+      }
```
Given that `returnToCache` was unintentionally ignored and partially fixed by https://github.com/aurelia/templating/commit/b4abe6e1bdd2e3fd66d2f566be8056f9202d273b#diff-903d639877fa6d7ca544fb24e9895617, then I think this commit is safe. @EisenbergEffect

dannyBies:
I'm not sure if it matters but I also noticed a difference in ` _projectionRemoveAll ` compared to other `remove...` methods.

The other functions possibly call detached() first and then returnToCache() while  ` _projectionRemoveAll ` can only do one or the other because it's an if/else if instead of if/if. I wasn't sure if this behaviour was intended or not so I kept it the same. If the fix should be two seperated if's then we should also swap the function calls so we call detached() before returnToCache() to keep it the same as in the other `remove...` functions.

The other `remove...` functions look like this:
```
if (this.isAttached) {
    view.detached();
}
if (returnToCache) {
    view.returnToCache();
}
```

And the ` _projectionRemoveAll ` 
```
if (returnToCache) {
    children[i].returnToCache();
} else if (this.isAttached) {
    children[i].detached();	
}
```

bigopon:
@dannyBies `returnToCache` does cover `isAttached` 
https://github.com/aurelia/templating/blob/b4abe6e1bdd2e3fd66d2f566be8056f9202d273b/src/view-factory.js#L398-L411

Notice `view.detached()` is called regardless of `ViewSlot` `isAttached` state

dannyBies:
@bigopon  That makes sense, thanks for looking into it!

Is it possible to get this merged & released soon? My project is leaking a lot of memory and I would like to fix this asap for our users.

jmezach:
Is there anything blocking this from being merged and released?

bigopon:
@jmezach @dannyBies This will needs final review from @EisenbergEffect

EisenbergEffect:
Thanks for working on this!

jmezach:
FYI: By adopting the latest 1.8.3 version, we've seen a dramatic decrease in memory leaks in our real world application. Before this version this would be the memory usage after clicking around in our application:

screenshot 2018-09-07 12 34 03

With the fix in place, we now see this:

image

So some signification improvements here. Note that we also changed some of our own code, which also accounts for some of the memory usage, but still.

bigopon:
@jmezach @dannyBies That is nice to see. Glad that this got resolved. I think it has taken few iterations of fixes around this. Thanks for your help 👍

PR #640, opened 2018-09-06
fix(view): skip view distribution on slot fallback fix #639

### LINKED ISSUE #639: Nested slot creates fallback even with override

* **Library Version:**
1.8.2

* **Browser:**
all

* **Language:**
all

**Current behavior:**
If we have nested slots the *fallback* content, for the nested one, gets created(`ShadowSlot.prototype.renderFallbackContent` gets called) even when there is an override provided for that slot. After that the fallback view is bound and unbound, it is never attached.

* **What is the expected behavior?**
[Repro](https://stackblitz.com/edit/aurelia-javascript-lsxduk?file=src%2Fapp.html) - I'm expecting that when I have nested `<slot>`s and target the nested one, its fallback content is *not* created, as is the case with unnested `<slot>`s.

* **What is the motivation / use case for changing the behavior?**
1. Unnecessary processing and allocation of resources since `<slot>`s are static, so the created fallback `View` will never be used. 
2. In certain cases, when we have custom elements/attributes in the nested slot fallback, the unexpected binding/unbinding of the fallback `View` can lead to errors since unexpected(incorrect) data will get bound to them.

**Observation**
For nested slots with override for the nested one, it seems that when `ShadowSlot.prototype.renderFallbackContent` is called for the nested `<slot>`, it is done based on  `needsFallbackRendering` evaluating to `true` because:
- at this time `projections` is still `0`(for the nested one)
- the call to `ShadowDOM.distributeView` => `ShadowDOM.distributeNodes`, which will modify `projections` correctly is still to happen
- the call to `ShadowDOM.distributeView` is done during binding of the `View`

StrahilKazlachev:
`ShadowSlot` and `PassThroughSlot` have *custom* logic that calls [`ShadowDOM.distributeNodes`](https://github.com/aurelia/templating/blob/1.8.2/src/shadow-dom.js#L300)/[`ShadowDOM.distributeView`](https://github.com/aurelia/templating/blob/1.8.2/src/shadow-dom.js#L48) after binding the fallback `View`. Because of that `ShadowDOM.distributeView` should *not* be called in [`View.prototype.bind`](https://github.com/aurelia/templating/blob/1.8.2/src/view.js#L170) for fallback `View`s. Calling it results in the creation and binding of the *fallback* `View` for *tested* `<slots>`s *even* if override is provided for *that* nested `<slot>`. 
@EisenbergEffect @bigopon

### DIFF
--- a/src/shadow-dom.js
+++ b/src/shadow-dom.js
@@ -40,6 +40,7 @@ export class PassThroughSlot {
   renderFallbackContent(view, nodes, projectionSource, index) {
     if (this.contentView === null) {
       this.contentView = this.fallbackFactory.create(this.ownerView.container);
+      this.contentView._isSlotFallback = true;
       this.contentView.bind(this.ownerView.bindingContext, this.ownerView.overrideContext);
 
       let slots = Object.create(null);
@@ -277,6 +278,7 @@ export class ShadowSlot {
   renderFallbackContent(view, nodes, projectionSource, index) {
     if (this.contentView === null) {
       this.contentView = this.fallbackFactory.create(this.ownerView.container);
+      this.contentView._isSlotFallback = true;
       this.contentView.bind(this.ownerView.bindingContext, this.ownerView.overrideContext);
       this.contentView.insertNodesBefore(this.anchor);
     }
--- a/src/view.js
+++ b/src/view.js
@@ -88,6 +88,7 @@ export class View {
     this.viewModelScope = null;
     this.animatableElement = undefined;
     this._isUserControlled = false;
+    this._isSlotFallback = false;
     this.contentView = null;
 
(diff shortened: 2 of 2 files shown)


EisenbergEffect:
Seems reasonable but I get nervous when changing code in here because there aren't really tests and it's rather complicated and a little shaky code.

bigopon:
@StrahilKazlachev That looks nice to me. We can test with element first, and then later will test with slot scenarios where it involves only text node. For unit test, I think if you already have pin pointed the issue, maybe a test for that specific part only if possible would be better, @EisenbergEffect 's call

EisenbergEffect:
I'd be very happy with some more tests like this, yes :)

PR #641, opened 2018-09-06
fix(StaticViewStrategy): set formal ".moduleId"

StrahilKazlachev:
Add *formal* `.moduleId` - `'undefined'`, so that one does *not* get [assigned](https://github.com/aurelia/templating/blob/1.8.2/src/html-behavior.js#L207) later on.
@bigopon


bigopon:
This will save probably a lot of work, because `Origin.get()` will iterate all modules. lgtm @EisenbergEffect

PR #643, opened 2018-09-20
fix(StaticViewStrategy): use DOM.Element instead of HTMLTemplateElement

bigopon:
@EisenbergEffect I missed this part. Thanks to @gimerstedt for pointing this out

### DIFF
--- a/src/view-strategy.js
+++ b/src/view-strategy.js
@@ -271,7 +271,7 @@ export class StaticViewStrategy {
   factory: ViewFactory;
 
   constructor(config: string | HTMLTemplateElement | IStaticViewConfig) {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Will this work correctly in browsers that don't have a native template element?

bigopon:
It will work in the same way however it is before this. We have a pending fix related to browser that doesn't support it natively. I remember @StrahilKazlachev did that PR.

For the class itsef, if it's a string and not start with `<template/>`, error will still be thrown like usual. If someone passes in an element that is not a template element, error will be throw later when it assume it's a config object. I think it's safe.

PR #644, opened 2018-09-21
fix(StaticViewStrategy): check for null template before compiling

StrahilKazlachev:
When `.template` is `null` - `@view({template: null, dependencies: [SomeResource]})`, act as `NoViewStrategy` - process the dependencies and return `null` for factory.
`@view(null)` already works as `@noView`.

### DIFF
--- a/src/view-strategy.js
+++ b/src/view-strategy.js
@@ -121,7 +121,7 @@ export class NoViewStrategy {
   * @param dependencies A list of view resource dependencies of this view.
   * @param dependencyBaseUrl The base url for the view dependencies.
   */
-  constructor(dependencies?: Array<string|Function|Object>, dependencyBaseUrl?: string) {
+  constructor(dependencies?: Array<string | Function | Object>, dependencyBaseUrl?: string) {
     this.dependencies = dependencies || null;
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
I'd like for @bigopon to due a review on this just to make sure it's good since he's been in this code a lot recently.

bigopon:
lgtm, though I'm not sure when this could be useful, maybe because I haven't had that usage yet. Would be nice if @StrahilKazlachev could give some example

StrahilKazlachev:
I did convert most of my codebase to use static resources, all except the router views. Now I'm working on a custom element with `@noView` and `@processContent(false)`. The element is processing and compiling the content itself.

Did ask @bigopon if he can think of a way to register the resources for this element without directly interacting with the `ViewResources`. He did suggest trying `@view({template: null, dependencies: [...]})` but it currently errors.
The simple form `@view(null)` already handles `null`.

StrahilKazlachev:
It's something like a dedicated `TemplateController`*-ish* custom element.

bigopon:
> It's something like a dedicated `TemplateController`-ish custom element.

Nice usage 👍

PR #647, opened 2018-10-08
Added a closing tag to the let elements

tomschluter:
It appears that the examples of the usage of the let element don't seem to work without a closing tag. Unfortunately a self closing tag doesn't seem to work either. So I have added closing tags to the let elements for now.

### DIFF
--- a/doc/article/en-US/templating-custom-elements.md
+++ b/doc/article/en-US/templating-custom-elements.md
@@ -305,7 +305,7 @@ Either using interpolation:
 
 <code-listing heading="Declarative Computed via Interpolation">
   <source-code lang="HTML">
-    <let full-name="${firstName} ${lastName}">
+    <let full-name="${firstName} ${lastName}"></let>
     <div>
       First name:
       <input value.bind="firstName" />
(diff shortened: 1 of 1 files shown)


huochunpeng:
Note self-closing is xml syntax, not HTML5 syntax.

In HTML5, `<any-tag />` is treated as `<any-tag>`, the `/` is ignored totally. HTML5 only defined few tags without need of closing like `<input>`, `<br>`, `<hr>`, ``, `<embed>`, `<col>`, `<param>`, `<source>`, `<track>` and in html head `<meta>`, `<base>`. I might missed some.

When you write `<input />` in HTML5, it still works because HTML5 sees it as `<input>`. This is the same reason why `<let />` didn't work as intended.

bigopon:
Thanks @tomschluter @huochunpeng.
Lgtm @EisenbergEffect. Sry i missed this

huochunpeng:
@bigopon note not only the <let> is incorrect, the <input> also get unnecessary `/>` ending which is bad example of HTML5. The blog post is affected too.

@tomschluter do you mind to clean up all `<input>` tags too in this PR?

PR #650, opened 2018-10-12
fix(Let): mark as target after creating instrutions

bigopon:
On gitter:
> I'm seeing a few warning messages as a result of using the new let functionality. Not an issue - but is this pointing at a possible issue?
WARN Detected string literal in let bindings. Did you mean "auTargetId.bind=65" or "auTargetId=${65}" ?
from this code in a custom element.
<let interested.bind="important ? 'Important' : 'Interested'"></let>

Atm, `<let/>` gets transformed into an instruction target too early. Should delay until after binding language has finished creating the instructions. Thanks 
@EisenbergEffect 

Thanks to @MylesPenlington

### DIFF
--- a/src/view-compiler.js
+++ b/src/view-compiler.js
@@ -320,13 +320,13 @@ export class ViewCompiler {
       // and the binding language has an implementation for it
       // This is an backward compat move
       if (tagName === 'let' && !type && bindingLanguage.createLetExpressions !== defaultLetHandler) {
-        auTargetID = makeIntoInstructionTarget(node);
         instructions[auTargetID] = TargetInstruction.letElement(
           bindingLanguage.createLetExpressions(
             resources,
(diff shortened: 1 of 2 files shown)


MylesPenlington:
The below is sufficient in a custom element to cause the issue - it seems like any reference of a let and its usage is sufficient. e.g. a html only custom element is sufficient to cause the issue. By the way I ended up doing this because it appears if.bind does not work on <slot if.bind="expression"> at the moment in a custom element - not certain if this should work.

```
<template>
   <let display.bind="isVisible ? 'inline': 'none'"></let>
   <div css="display: ${display};">
            <slot>Missing content</slot>
   </div>
</template>

export class IndicatorCustomElement {
    public isVisible: boolean = false;
}
```

PR #651, opened 2018-10-26
fix(Let): properly create instruction Id

bigopon:
Current behavior is having all calls in a single statement, this leads to a bug where instruction id cannot be properly created before creating instruction. Thanks to @graycrow for pointing this out.

cc @EisenbergEffect


EisenbergEffect:
ok, well I can get these fixes out now since it's not working anyways 🤣 Thanks for jumping on this!

Issue #657, opened 2019-02-28
ElementEvents.subscribe set useCapture to true by default, which is not standard


sebastien-roch:
**I'm submitting a bug report**

* **Library Version:**
master

* **Language:**
all

**Current behavior:**
When using `EventElements.subscribe`, the listener will be set with `useCapture = true`, which is not the standard behavior of the native `addEventListener()` function and thus confuses the developer.

* **What is the expected behavior?**
`EventElements.subscribe` (and subscribeOnce) should use set `useCapture` to `false` by default.

* **What is the motivation / use case for changing the behavior?**
Mimic native API of the browser: addEventListener would set `useCapture` to `false` if not passed in arguments.

bigopon:
@sebastien-roch I've been hit by this as well. It's unfortunate, but we can only fix this in a major release of templating.

cc @EisenbergEffect

EisenbergEffect:
Sadly, we can't change it at this point because it could potentially break existing apps. What if we had some setting that could be set to get the desired behavior though, in a non-breaking way?

bigopon:
@EisenbergEffect We cna do that via a static field on `ElementEvents` class, say `defaultListenerOptions`

EisenbergEffect:
That works for me. @sebastien-roch Would you be interested in adding this property and updating the implementation to use it? I think it would make a fine contribution and we'd be very happy to work with you to get your PR in. Thoughts?

sebastien-roch:
I can do that. Towards which branch should I target my pull request having this `defaultListenerOptions`? I guess I have to create a second PR for a future breaking release where we just set `capture` to `false` by default.

bigopon:
@sebastien-roch Thanks for that. It should be `develop` branch

PR #658, opened 2019-03-18
fix(ElementEvents): ElementEvents.defaultListenerOptions allows to set default options when subscribing to events

sebastien-roch:
ElementEvents.subscribe will set capture to true be default if not explicitely passed in argument. This is not the standard behavior of the native addEventListener() and confuses developers. This new options let you define the default behavior.

Fixes issue #657

### DIFF
--- a/src/element-events.js
+++ b/src/element-events.js
@@ -13,6 +13,8 @@ interface EventHandler {
  * @param element
  */
 export class ElementEvents {
+  static defaultListenerOptions: boolean | AddEventListenerOptions = { capture: true };
+
   constructor(element: EventTarget) {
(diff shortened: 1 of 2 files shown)


src/element-events.js
```diff
@@ -13,6 +13,8 @@ interface EventHandler {
  * @param element
  */
 export class ElementEvents {
+  static defaultListenerOptions: boolean | AddEventListenerOptions = { capture: true };
```
bigopon: Nice. But this is unfortunately a breaking change as well, as it doesn't work for IE users. I think it should be 
`defaultListenerOptions = true`

PR #659, opened 2019-03-18
fix(ElementEvents): ElementEvents.defaultListenerOptions allows to set default options when subscribing to events

sebastien-roch:
ElementEvents.subscribe will set capture to true be default if not explicitely passed in argument. This is not the standard behavior of the native addEventListener() and confuses developers. This new options let you define the default behavior.

Fixes issue #657

### DIFF
--- a/src/element-events.js
+++ b/src/element-events.js
@@ -55,7 +55,7 @@ export class ElementEvents {
    */
   subscribe(eventName: string, handler: Function, captureOrOptions?: boolean | AddEventListenerOptions): EventHandler {
     if (typeof handler === 'function') {
-      if (typeof captureOrOptions === 'undefined') {
+      if (captureOrOptions === undefined) {
         captureOrOptions = ElementEvents.defaultListenerOptions;
       }
       const eventHandler = new EventHandlerImpl(this, eventName, handler, captureOrOptions, false);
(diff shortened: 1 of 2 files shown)


src/element-events.js
```diff
@@ -13,6 +13,8 @@ interface EventHandler {
  * @param element
  */
 export class ElementEvents {
+  static defaultListenerOptions: boolean | AddEventListenerOptions = { capture: true };
```
bigopon: I commented on the other PR that was closed. So i'll copy paste here

> Nice. But this is unfortunately a breaking change as well, as it doesn't work for IE users. I think it should be
defaultListenerOptions = true

sebastien-roch: You're right, fixed that

bigopon: Nice!

sebastien-roch:
Sorry I just had to force-push last commits to fix the author e-mail address.

src/element-events.js
```diff
@@ -51,8 +53,11 @@ export class ElementEvents {
    * Adds and Event Listener on the context element.
    * @return Returns the eventHandler containing a dispose method
    */
-  subscribe(eventName: string, handler: Function, captureOrOptions?: boolean = true): EventHandler {
+  subscribe(eventName: string, handler: Function, captureOrOptions?: boolean | AddEventListenerOptions): EventHandler {
     if (typeof handler === 'function') {
+      if (typeof captureOrOptions === 'undefined') {
```
bigopon: I'll repeat what @jdanyow said to me about this case: the rest of the code base doesn't care about crazy people changing default value of `undefined`, so lets just compare `captureOrOptions === undefined`

sebastien-roch: ok changed that

Issue #670, opened 2019-07-11
Failing karma tests report incorrect line numbers in console stacktrace


baerrach:
**I'm submitting a bug report**

* **Library Version:**
aurelia-templating: "1.10.3"

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
10.15.3

* **NPM Version:**
6.4.1

* **JSPM OR Webpack AND Version**
JSPM 0.16.53

**Current behavior:**
Create a new test case and force it to fail.

```
expect('force a fail).toEqual('failure);
```

Not the line numbers with your unit test and the stack trace do not match.

**Expected/desired behavior:**

* **What is the expected behavior?**
The line number match the source code.

PR #671, opened 2019-07-11
fix(karma): add retainLines to babelPreprocessor

baerrach:
When a test fails under Karma the console stacktrace contains incorrect line
numbers making debugging impossible.

Applying fix manually as described in https://github.com/chriscasola/karma/commit/67d1cfe5bc4a85c039bb4d8af36b366879aa6cc0

closes aurelia/templating#670

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -36,6 +36,7 @@ module.exports = function(config) {
     'babelPreprocessor': {
       options: {
         sourceMap: 'inline',
+        retainLines: true,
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Checking briefly with @fkleuver to see if he’s familiar with this setting before I merge.

fkleuver:
I don't have experience with it but I don't see any harm in it. The CI jobs succeed so if it helps people then let's get it in

PR #677, opened 2020-01-19
Fix slot leak

bigopon:
Fixes aurelia/templating-resources/issues/392

Thanks to @thomas-darling for this fix.

---

Currently, `ShadowSlot` keeps track of what `Node` has been projected through it, via an internal `children` array but it doesn't untrack(remove) those tracked `Node` from the array when the owning views of those `Node`s get removed. Fix this by removing them in `ShadowSlot.prototype.removeView` and `ShadowSlot.prototype.removeAll`

Also revamp testing setup so that it can remove the usage of jspm for testing, while keeping the build process intact. Now we have:
```shell
npm run test
npm run test:watch
npm run test:debugger
```
with an option to go full integration tests to have more solid infra for tricky parth such as Shadow DOM emulation.

@EisenbergEffect @fkleuver @StrahilKazlachev


bigopon:
@fkleuver @EisenbergEffect @StrahilKazlachev  ready for review

StrahilKazlachev:
@bigopon I've not been active for some time, so are dist files added now with the PRs?

bigopon:
@StrahilKazlachev its because the target branch is develop and the pr is based on master. The commits that contain those dist should ideally ve in develop already. Cc @fkleuver

bigopon:
resolved by #678

PR #678, opened 2020-01-22
fix(shadow-dom): prevent slot leak

bigopon:
Copied from #677 

----

Fixes aurelia/templating-resources/issues/392

Thanks to @thomas-darling for this fix.

---

Currently, `ShadowSlot` keeps track of what `Node` has been projected through it, via an internal `children` array but it doesn't untrack(remove) those tracked `Node` from the array when the owning views of those `Node`s get removed. Fix this by removing them in `ShadowSlot.prototype.removeView` and `ShadowSlot.prototype.removeAll`

Also revamp testing setup so that it can remove the usage of jspm for testing, while keeping the build process intact. Now we have:
```shell
npm run test
npm run test:watch
npm run test:debugger
```
with an option to go full integration tests to have more solid infra for tricky parth such as Shadow DOM emulation.

@EisenbergEffect @fkleuver @StrahilKazlachev


EisenbergEffect:
Thank you @bigopon  and @thomas-darling !

@thomas-darling were you able to confirm that this works well in your app without any breaks?
@bigopon Thanks for the extra work you put into improving things, besides the original fix. Is this all ready to merge?

bigopon:
Yes it is

EisenbergEffect:
Ok, lets see if our dev branch build and merge CI works with the new stuff. Here goes...

thomas-darling:
@EisenbergEffect
So far it works perfectly in our app - nothing breaking, and it appears to have fixed the leaks :-)

Do you have an ETA for a release that includes this fix?
I have some other apps in production that, although less critical, could use this fix too :-)

EisenbergEffect:
I should be able to get the release out within the next few days.

PR #685, opened 2020-05-17
Fix: ensure child observation work properly with native & emulation for content projection 

### LINKED ISSUE #675: @child() decorator does not work properly in combination with if.bind

* **Library Version:**
1.10.2

* **Operating System:**
Windows 10

* **Node Version:**
10.16.3

* **NPM Version:**
6.12.0

* **JSPM OR Webpack AND Version**
JSPM 0.16.53

* **Browser:**
all

* **Language:**
all

**Current behavior:**
Let's say we have an app, which uses the component `simple-component`. This component is placed inside container which is added/removed conditionally. This component uses `@child()` decorator to reference an `input` element.

**app.html**
```html
<template>
    <require from="./simple-component/simple-component"></require>
    <button click.delegate="toggleContainers()">Toggle Containers</button>
    <div class="container-1"
         if.bind="showMainContainer">
        <simple-component>
            <input type="text" id="test-input">
        </simple-component>
    </div>
    <div class="container-2" else>Placeholder container</div>
</template>
```

**app.ts**
```typescript
export class App {
    showMainContainer: boolean = true;

    toggleContainers() {
        this.showMainContainer = !this.showMainContainer;
    }
}
```

**simple-component.html**
```html
<template>
    <slot></slot>
    <span>Some Input Suffix</span>
</template>
```

**simple-component.ts**
```typescript
import {child} from 'aurelia-templating';

export class SimpleComponent {
    @child('input') myInput: any;

    attached() {
        console.log('myInput: ', this.myInput);
    }
}
```

On initial load, reference in `myInput` property is correct. But if we click "Toggle Containers" button two times (to remove and add `cointainer-1` element to the DOM again) then `myInput` property points to `null`. I noticed that current behaviour is caused by this commit: https://github.com/aurelia/templating/commit/7989015. In previous versions of this plugin everything worked fine, even in scenario described above.

* **What is the expected behavior?**

Component's property decorated with `@child` decorator should reference appropriate element even if component's parents are added or removed conditionally.

bigopon:
At the moment, in `unbind()` of `ChildObserverBinder`, it's always unset the value via
```ts
this.viewModel[this.property] = null;
```
This works for native shadow DOM, because it's always ensured to retrieve the value again during `bind()`, as it's safe to do so. Originally, we didn't have this, and it was my attempt of trying to do proper cleanup, and it resulted in this subtle bug between native & emulation.
To actually fix it, the above needs to be nested inside a native shadow DOM usage check:
```ts
if (this.useShadowDOM) {
  this.viewModel[this.property] = null;
}
``` 

For emulation, the content projection is done via an augmented view slot, and thus, can only work when content is actually projected through it, hence `.bind()`/`.unbind()` won't, and shouldn't affect it.

This PR also:
- guards change handler of child observation so that it's only called when the binder is bound.
- adds 11 integration tests for assertion different combination of `if`/`repeat` & `parent/child` to ensure that it works properly.

@EisenbergEffect 

Close #675

### DIFF
--- a/build/tasks/lint.js
+++ b/build/tasks/lint.js
@@ -6,5 +6,7 @@ gulp.task('lint', function() {
   return gulp.src(paths.source)
     .pipe(eslint())
     .pipe(eslint.format())
-    .pipe(eslint.failOnError());
+    // failOnError:     bail immediately on any error
+    // failAfterError:  aggregate all errors before bailing
+    .pipe(eslint.failAfterError());
 });
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -41,7 +41,7 @@ function(config) {
       performance: {
         hints: false,
       },
-      devtool: Array.isArray(browsers) && browsers.includes('ChromeDebugging') ? 'eval-source-map' : 'inline-source-map',
+      devtool: Array.isArray(browsers) && browsers.includes('ChromeDebugging') ? 'inline-source-map' : 'inline-source-map',
       module: {
         rules: [
           {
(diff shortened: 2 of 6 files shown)


EisenbergEffect:
Looks like you are still working on this so just let me know when you feel that it's ready to merge. @bigopon I'm also fine with you doing the merge and taking on the lead role around au1 templating. You clearly know more than anyone else now. It's been a while since I looked into the depths of this area of code myself, so I trust you more than I trust myself 😄

bigopon:
@EisenbergEffect All done 2 commits ago. Just this morning I realize the way linting process bails at the moment could easily lead to 2 - 3 follow up commits so I changed it to aggregating all linting errors before exiting. Ready to go any time.
With regards to taking the lead, sure I'll take care of that. Though it's good to have someone looking at this fresh.

Mostly it's the change from
```ts
  unbind() {
    if (this.viewHost.__childObserver__) {
      this.viewHost.__childObserver__.disconnect();
      this.viewHost.__childObserver__ = null;
      this.viewModel[this.property] = null;
    }
  }
```
to:
```ts
  unbind() {
    if (has observer) {
      // 1. remove this binder
      // 2. check if there's no binder on the observer, and remove the observer itself if 0 binder
      // 3. check if it's safe to unassign the value in the view model, via useShadowDOM property
    }
  }
```

PR #705, opened 2022-03-31
fix(dep): resolve vulnerable dep issue

bigopon:
The `templating` module is having dep dependency on `templating-resources` with the range `1.0.0`, upgrade this to `1.14.0` so that the security issue will not appear in the dep tree of consumer packages of this package. Also took the opportunity to refactor all code to TS, change the build script & ci. It should be much easier to maintain this package from now on.


bigopon:
This is part of the resolution for the issue reported at https://github.com/aurelia/framework/issues/992

